PR #31, opened 2024-05-26
feat: change tag data to correct tags array format

zriyansh:


Calm-Rock:
@zriyansh @YounixM is this PR helpful or can we close this ?

PR #196, opened 2024-06-04
fix: Remove code blocks from toc heading generation

harshbutfairx:
# Before
image

# After
image


ankit01-oss (approved):

contentlayer.config.ts
```diff
@@ -59,7 +59,17 @@ const computedFields: ComputedFields = {
...
+      const bodyWithoutCodeblocks = doc.body.raw.split('```').reduce((acc, ele, index) => {
+        // Every alternate block of text would be outside of the codeblock
+        if (index %2 === 0)  {
+          acc += ele;
+        }
+
+        return acc;
+      }, '');
+
+      const headings = Array.from(bodyWithoutCodeblocks.matchAll(regXHeader)).map(({ groups }) => {
```
YounixM, line 72: We can write this to be clear.  We should use regex.

YounixM:
```
      const regXCodeBlock = /```[\s\S]*?```/g;
      // Remove code blocks from the raw content
      const contentWithoutCodeBlocks = doc.body.raw.replace(regXCodeBlock, '');
```

harshbutfairx: Added the above changes.

ankit01-oss (approved):

PR #201, opened 2024-06-05
fix: use authors constant to render authors info

YounixM:


pranay01:
@YounixM where is the new place where author info should be added

cc @ankit01-oss might be helpful for your to understand how to add author infos

yuvraajsj18 (approved):

YounixM:
@pranay01 : all author info resides in the authors.json file in constants folder.

PR #208, opened 2024-06-06
fix: added note for fargate in self-host

Calm-Rock:


prashant-shahi (approved):
LGTM!

`public/search.json` file was also updated, is that expected?

Calm-Rock:
> LGTM!
> 
> `public/search.json` file was also updated, is that expected?

Yes

PR #210, opened 2024-06-07
Add docs for Histogram panel

srikanthccv:


srikanthccv:
@Calm-Rock @ankit01-oss please review.

Calm-Rock:
@srikanthccv can you please do 2 things:
- create a folder called **dashboards** under docs directory and move this doc under that folder.
- also for the images under the docs folder can you create a folder called **dashboards** and have another sub-folder called **panel-types** and put the images under that folder.

userguide folder is something we won't be using from now on for all the new docs. 
Also, I will be working on a PR for bifurcating the Dashboards and Querying Sections so this will be helpful there as well.

srikanthccv:
These are housekeeping changes. I would prefer to not get into such clean-up in this PR because I plan to share these docs with the user in the night and not delay. You can create a separate issue and assign it to me.

Calm-Rock (approved):

PR #266, opened 2024-06-14
Remove incorrect note about multiple channels

srikanthccv:
Setting up multiple channels doesn't send notifications to all channels. When creating an alert, the user should explicitly select the option to send to all channels.


srikanthccv:
@Calm-Rock please review

Calm-Rock (approved):

PR #274, opened 2024-06-15
Add instructions for OTel setup in Celery worker processes

srikanthccv:
Fixes https://github.com/SigNoz/engineering-pod/issues/1385
Fixes https://github.com/SigNoz/engineering-pod/issues/1340


data/docs/instrumentation/celery.mdx
```diff
@@ -0,0 +1,69 @@
...
+title: Celery Worker OpenTelemetry Setup
+description: Set up OpenTelemetry instrumentation in your Celery worker applications.
+---
+
+
+import InstrumentationFAQ from '@/components/shared/instrumentation-faq.md'
+
+
+This document contains instructions on how to set up OpenTelemetry instrumentation in your Celery worker applications.
+
```
Calm-Rock, line 13: Can we add an admonition mentioning that these instructions are valid for both SigNoz Cloud and SigNoz self-host users

srikanthccv: It would make sense to write that if there is a chance of confusion. This documentation mentions nothing about where one sends data. It is entirely about the shortcomings of the Celery library and how to work around it.

srikanthccv: Read this, for example, https://pymongo.readthedocs.io/en/stable/faq.html#is-pymongo-fork-safe. It talks about the issue of the library, the real database where it is talking to is indifferent.

srikanthccv: My line of reasoning is the same as why there is no need to write VM, docker, or k8s-specific sections here because they would be the same and don't need additional notes. If you think otherwise, please help me understand why.

Calm-Rock: That's a fair point. If the instructions are the same for all the environments the user will work on, then it doesn't make any sense to create different sections.


But I think it would be good to just explicitly mention that these instructions are valid for all environments. 
This will give users clarity, and they probably won't have to ask if this doc is valid for their environment.

srikanthccv: So does it mean that all the docs we have today (and whatever will be added in future) should explicitly mention that instructions work in every environment because that would give clarity and they don't have to ask if the doc is valid for their environment? I can add the instructions but I am trying to understand the framework that defines when to add them. For instance, the same should have been mentioned here https://github.com/SigNoz/signoz-web/pull/277

srikanthccv: I personally do not think it's needed but if we have a framework that we abide by I would gladly follow that because it shouldn't be about personal preferences or beliefs.

Calm-Rock: Yes, if we have a document that is valid for all environments, it should mention that it is valid for all environments.
I have seen instances where users do ask about instructions for their environments or request instructions for their particular environment (especially for Cloud).

Thanks for pointing out #277 - yes it should also have this mentioned. Will work on a PR for this ([issue here](https://github.com/SigNoz/signoz-web/issues/298))

This is an iterative process and will keep evolving - we don't have a very solid framework written down anywhere yet, so I am working on creating a Guide for documentation ([Issue](https://github.com/SigNoz/signoz-web/issues/262)) where I will list down the general things to keep in mind.

srikanthccv: > I have seen instances where users do ask about instructions for their environments or request instructions for their particular environment (especially for Cloud).

Please share these instances because the most important thing here is context. You have gotten such requests because there are already env-specific instructions for A, B, and C and missing for X, Y, and Z which they requested. My point here is that there is no concept of the env here and shouldn't be. I also shared an example but I can share more such docs from well-known public companies where there is no need to mention the env because the doc is about the issue with the library.

srikanthccv: Did you receive any questions on a doc that didn't mention about env if they work across all environments?

Calm-Rock: I can't find any such documentation where we don't have instructions segregated by environment. 
I understand your point. Merging this PR now.

Calm-Rock (approved):

PR #477, opened 2024-07-16
feat: added API to footer

Calm-Rock:


components/mainFooter.tsx
```diff
@@ -28,6 +28,9 @@ function Footer() {
                   </Link>
                   <ArrowUpRight size={16} />
                 </div>
+                <Link href="/api_reference/" className="mt-5 hover:underline">
+                  API
```
pranay01: @Calm-Rock Should we make it `SigNoz API Docs` ? Just API is not very clear

Calm-Rock: `SigNoz API` should be good since it's already under the Docs section.

pranay01 (approved):

PR #528, opened 2024-07-23
Updated link for security and compliance

ankit01-oss:


pranay01:
Do we not use - https://signoz.io/security/ anywhere else? Ideally this should also redirect to https://trust.signoz.io

This PR only replaces the footer link - but  https://signoz.io/security/  is still there

pranay01 (approved):

PR #591, opened 2024-08-03
fix: fixed the button name

Calm-Rock:


app/tag-data.json
```diff
@@ -1,7 +1,7 @@
 {
   "signoz": 27,
   "community": 6,
-  "prometheus": 48,
+  "prometheus": 60,
```
vikrantgupta25, line 4: are these intended ?

vikrantgupta25 (approved):

app/tag-data.json
Calm-Rock: These are autogenerated.

PR #624, opened 2024-08-13
docs: replace traefik's cli parameter name

jinrenjie:
Replace traefik's cli parameter name with the new version.


prashant-shahi (commented):
can you update the `image: traefik:v3.0.0-beta3` tag as well to the relavant ones?

jinrenjie:
@prashant-shahi Yes, I used the `latest` tag instead of the fixed version tag, maybe that's more universal!

prashant-shahi (approved):

prashant-shahi:
@jinrenjie yes, that should not be a problem. Thanks for the PR! :tada:

PR #800, opened 2024-09-20
fix: Typo Azure Blob Storage SideNav

konnta0:
SSIA


Calm-Rock (approved):
Thanks for the contribution @konnta0! You've got a sharp eye for details.

PR #807, opened 2024-09-24
update: updated vector logs docs

Calm-Rock:
This PR 
- Updates the structure of Vector logs docs
- Updates logic for tabs component so that it renders an mdx file even if we have a single tab. 
Currently, it throws an error when we try to have a single tab. Example of a single tab doc - [here](https://signoz-web-git-vector-docs-update-signoz.vercel.app/docs/logs-management/send-logs/vector-logs-to-signoz/)


components/Tabs.tsx
```diff
@@ -48,4 +48,4 @@ const Tabs = ({ children }) => {
   );
 };
 
-export default Tabs;
+export default Tabs;
```
YounixM, line 51: all imports that were previously using the `Tab` component should be updated to `Tabs`

ankit01-oss (approved):

PR #951, opened 2024-11-08
chore: use api.signoz.cloud instead of signup

prashant-shahi:
#### Chores
- use api.signoz.cloud instead of signup


therealpandey:
This has already been removed!
@YounixM if you want to delete this older code or not.

PR #1001, opened 2024-12-02
chore: resolve sentry react to v7.19.1

YounixM:
To Fix: https://github.com/SigNoz/signoz-web/security/dependabot/17

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "@nextui-org/react": "^2.4.2",
     "@opentelemetry/api": "^1.9.0",
     "@signozhq/design-tokens": "^0.0.8",
-    "@stoplight/elements": "^8.3.3",
+    "@stoplight/elements": "^8.5.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -90,7 +90,9 @@
     "typescript": "^5.1.3"
   },
   "resolutions": {
-    "tar": "6.2.1"
+    "tar": "6.2.1",
+    "@sentry/react": "^7.19.1",
```
SagarRajput-7, line 94: @YounixM 
on the signoz repo we have - `"@sentry/react": "8.41.0",` and also `^` does not cater -upgrade to 8.x.x, since major version changes often introduce breaking changes. 

approving if this is expected and according to the other dependencies and compatibility of this repo

SagarRajput-7 (approved):

package.json
YounixM: To Fix: https://github.com/SigNoz/signoz-web/security/dependabot/17

PR #1132, opened 2025-01-28
feat: new nav dropdown

sawhil:
Issue : https://github.com/SigNoz/engineering-pod/issues/2150


components/TopNav/TopNav.tsx
ahmadshaheer, line 1: The hover effect seems different. is this expected? 

image

image

ahmadshaheer, line 1: Is the customer stories intended to be only for Product menu or for resources as well?

image

components/TopNav/TopNav.tsx
```diff
@@ -263,109 +280,230 @@ export default function TopNav() {
...
+                                  <span>{comparisionItem.name}</span>{' '}
+                                  <ArrowRight
+                                    className="opacity-0 group-hover:opacity-100"
+                                    size={14}
+                                  />
+                                </Link>
+                              ))}
+                            </div>
+                          </div>
+                        </div>
```
ahmadshaheer: can you extract this into a separate component? since the same code has been repeated on line 457 as well.

components/TopNav/TopNav.tsx
```diff
@@ -263,109 +280,230 @@ export default function TopNav() {
...
+                            Explore
+                          </div>
+                          <div className="grid grid-cols-1 gap-x-[42px] gap-y-[36px]">
+                            {resourcesDropdownItems.explore.map((item) => (
+                              <Link
+                                href={item.url}
+                                className="group flex h-auto items-center gap-4"
+                                key={item.key}
+                              >
+                                <div>
```
ahmadshaheer, line 475: nit: do we need this extra div?

components/TopNav/TopNav.tsx
```diff
@@ -263,109 +280,230 @@ export default function TopNav() {
...
+                            Learn
+                          </div>
+                          <div className="grid grid-cols-1 gap-x-[42px] gap-y-[36px]">
+                            {resourcesDropdownItems.learn.map((item) => (
+                              <Link
+                                href={item.url}
+                                className="group flex h-auto items-center gap-4"
+                                key={item.key}
+                              >
+                                <div>
```
ahmadshaheer, line 443: nit: do we need this extra div?

ahmadshaheer (changes requested):

components/TopNav/TopNav.tsx
sawhil: Yes

sawhil: As of now for both, but Shuvanshu will add something else later in the resources part.

GeekBoySupreme:
@sawhil there is a small overlap of divs in the time it takes for one to fade out and the other to open. Instead of handling them as two separate divs? Can we treat them as one div with 2 separate views/properties, if the hover state changes from `Product` to `Resources` or vice versa while the dropdown is open, it can simply morph to the other view instead of closing and opening another div. WDYT?

https://github.com/user-attachments/assets/5351a589-1ad4-4694-ab8d-aee0ee0ee1a5

sawhil:
> @sawhil there is a small overlap of divs in the time it takes for one to fade out and the other to open. Instead of handling them as two separate divs? Can we treat them as one div with 2 separate views/properties, if the hover state changes from `Product` to `Resources` or vice versa while the dropdown is open, it can simply morph to the other view instead of closing and opening another div. WDYT?
> 
>  Screen.Recording.2025-01-29.at.11.06.54.AM.mov

Yeah that is a good idea, but how to do you want to transition the content ?

What kind of animation you want for this transition. Otherwise it will be quite jittery.

Need to check on the positioning as well for this.

Can you give me an example of the same ? @GeekBoySupreme

components/TopNav/TopNav.tsx
sawhil: Removed this from other place as requirement changed from product.

components/TopNav/TopNav.tsx
sawhil: Yeah @ahmadshaheer We need it as removing it is putting it in row flex.

components/TopNav/TopNav.tsx
sawhil: Answered below

ankit01-oss (approved):

components/TopNav/TopNav.tsx
```diff
@@ -208,6 +226,15 @@ export default function TopNav() {
     const id = setTimeout(() => setIsOpenResources(false), delay)
     setTimeoutIdResources(id)
   }
+
+  const handleProductDropdownClick = () => {
+    setIsOpen(false)
```
YounixM, line 232: NIT : maintain consistency in naming functions

components/TopNav/TopNav.tsx
```diff
@@ -263,109 +289,190 @@ export default function TopNav() {
...
+                              >
+                                <div>
+                                  <div className="flex flex-row items-center gap-1">
+                                    <span>{item.name}</span>{' '}
+                                    <ArrowRight
+                                      size={14}
+                                      className="opacity-0 group-hover:opacity-100"
+                                    />
+                                  </div>
+                                  <div className="line-clamp-2 max-w-[274px] text-xs text-[#C0C1C3]  group-hover:text-[#FFF]">
```
YounixM: @sawhil : Check if we can use design tokens for the hex code

YounixM (commented):
The brainfish svg has white background corners visible, please check.

sawhil:
> The brainfish svg has white background corners visible, please check.

It seems to be on github preview only. Checked on the signoz-web it's not showing any white corners and also checked the SVG preview through a previewer.

components/TopNav/TopNav.tsx
sawhil: The state setter was already present should i rename the state and setter ?

components/TopNav/TopNav.tsx
```diff
@@ -347,17 +354,23 @@ export default function TopNav() {
                                   width={20}
                                   height={20}
                                 />
-                                <div className="font-inter line-clamp-2 max-w-[274px] text-[#C0C1C3] group-hover:text-[#fff]">
+                                <div
+                                  className={`font-inter line-clamp-2 max-w-[274px] text-[${Color.TEXT_VANILLA_400}] group-hover:text-[#fff]`}
```
ahmadshaheer, line 358: we can change #fff to Color.BG_VANILLA_100 (here and in other instances)

package.json
```diff
@@ -21,7 +21,7 @@
     "@next/third-parties": "^14.2.3",
     "@nextui-org/react": "^2.4.2",
     "@opentelemetry/api": "^1.9.0",
-    "@signozhq/design-tokens": "^0.0.8",
+    "@signozhq/design-tokens": "1.1.4",
```
ahmadshaheer, line 24: can you please confirm that upgrading this won't affect any existing usage of design tokens?

ahmadshaheer (commented):
minor comments. rest looks good to me

package.json
sawhil: Apart from Tailwind css config we aren't importing it anywhere.

components/TopNav/TopNav.tsx
sawhil: Tried it, but it didn't work so left it there

YounixM (approved):

components/TopNav/TopNav.tsx
ahmadshaheer: it should have worked, I guess, because you're already using `text-[${Color.TEXT_VANILLA_400}]`.

but no worries. will approve the PR. can you please take a look at why it doesn't work and maybe make the change in another PR?

ahmadshaheer (approved):

PR #1175, opened 2025-02-12
chore(license): update the signup apis to consume the new structure

vikrantgupta25:
- update the signup API to handle the new structure


vikrantgupta25:
contributes to - https://github.com/SigNoz/platform-pod/issues/450

PR #1221, opened 2025-03-06
fix: update instrumentation docs for Next.js on Vercel

kakadiadarpan:
`OTLPTraceExporter` uses `XMLHTTPRequest` under the hood which is incompatible with Vercel Edge runtime and should only be used with `Node.js` runtime. We recommend using `OTLPHttpJsonTraceExporter` from `@vercel/otel` instead for better compatibility with all the runtimes.


makeavish (approved):
Thanks for the PR!

PR #1300, opened 2025-03-31
chore: remove mixpanel and posthog changes

YounixM:


tsconfig.tsbuildinfo
SagarRajput-7, line 1: do we need it? or we can .git ignore?

SagarRajput-7 (approved):

Issue #1021, opened 2024-12-05
[EPIC] Frontend bugs/features in documentation


Calm-Rock:
Frontend issues related to docs

- [x] https://github.com/SigNoz/signoz-web/issues/571
- [x] https://github.com/SigNoz/signoz-web/issues/572
- [ ] https://github.com/SigNoz/signoz-web/issues/1020

YounixM:
@sawhil : Please share the updates on this

sawhil:
Handed the part 2 to @Calm-Rock (Chitransh).

PR #1314, opened 2025-04-07
fix: added dynamic tab based url query params

sawhil:


Calm-Rock:
https://github.com/SigNoz/signoz.io/issues/1021 - relevant issue.

PR #1350, opened 2025-04-18
Linux install guide improvment

gfelot:
Here my own version of the vanilla install of SigNoz on linux.


Calm-Rock:
@Nageshbansal PTAL if this PR is still relevant.

PR #1372, opened 2025-04-26
fix(search-modal): Added a fix for search modal in mobile screen

ahrefabhi:
### Added fix for search modal not rendering properly on mobile device.

- Previously
Screenshot 2025-04-27 at 1 54 32 AM

- Now
Screenshot 2025-04-27 at 1 56 21 AM

### DIFF
--- a/css/global.css
+++ b/css/global.css
@@ -239,6 +239,10 @@ body::-webkit-scrollbar-thumb {
   padding: 24px;
 }
 
+body #trieve-search-modal {
+  top: calc(40vh - 225px);
+}
+
 body #trieve-search-modal.dark .item.start-chat {
   @apply bg-signoz_sakura-600/10;
 }


css/global.css
```diff
@@ -239,6 +239,10 @@ body::-webkit-scrollbar-thumb {
   padding: 24px;
 }
 
+body #trieve-search-modal {
+  top: calc(40vh - 225px);
```
aks07, line 243: Curious how we came up with this calculation. What if we simply add
`top: 25%` or `top: 50%`

aks07 (changes requested):
Left a comment.

css/global.css
ahrefabhi: trieve-search-component internally adds the above css, but it was only done for desktop, because of which it started breaking on mobile.

aks07 (approved):

PR #1548, opened 2025-06-10
UA Tracking for Bots

yuvraajsj18:
- Enhanced the BOT_USER_AGENT_PATTERNS array to include specific AI data scrapers, user-triggered AI assistants, and undocumented AI agents.
- Improved categorization of bot detection patterns for better clarity and functionality in identifying various AI-related user agents.


utils/logEvent.ts
```diff
@@ -26,3 +26,168 @@ export const logEvent = async (payload: LogEventPayload) => {
...
+      return { isBot: true, botType: 'phantomjs', reason: 'phantom_detected' }
+    }
+
+    // Check for unusual screen dimensions (common in headless browsers)
+    if (window.screen && (window.screen.width === 0 || window.screen.height === 0)) {
+      return { isBot: true, botType: 'headless_browser', reason: 'zero_screen_dimensions' }
+    }
+  } catch (error) {
+    // If we can't access these properties, might be a bot
+    return { isBot: true, botType: 'restricted_access', reason: 'property_access_error' }
```
amlannandy, line 189: Will it always throw an error here if we can't access these properties?

middleware.ts
```diff
@@ -1,24 +1,99 @@
 import { NextResponse } from 'next/server'
 import type { NextRequest } from 'next/server'
+import { waitUntil } from '@vercel/functions'
 import { v4 as uuidv4 } from 'uuid'
+import { detectBotFromUserAgent, logEventServerSide } from './utils/logEvent'
+
+// Extract OS from user agent (server-side version)
+const getOSFromUserAgent = (userAgent: string): string => {
```
amlannandy, line 8: In this event, I asked chatgpt to scrape through the alerts doc. Is it expected to not have the `Browser` and `Device` filled here? Just wanted to confirm.

Screenshot 2025-08-13 at 3 07 07 PM

utils/logEvent.ts
ahmadshaheer: No, it won't throw an error. if we can't access any of the above properties, it'll return the object `{isBot: true, ...}`

middleware.ts
ahmadshaheer: Yes, @amlannandy. this is expected. the AI bots don't use a real browser, therefore the fields are empty.

ahmadshaheer (approved):

amlannandy (approved):

PR #1609, opened 2025-06-26
Update Quarkus guide to avoid incorrect suggestions

### LINKED ISSUE #8355: Quarkus guide is broken
This guide: https://signoz.io/docs/instrumentation/opentelemetry-quarkus/ is broken for a few reasons:
1. The `otlp` exporter for Quarkus does not work and is not intended to (https://github.com/quarkusio/quarkus/issues/48570)
2. The `http/protobuf` setup does not seem to work with Quarkus. It may be a SigNoz side issue because the trace is leaving the application just fine, but SigNoz never seems to ingest it. gRPC is the default Quarkus option anyways and works perfectly fine with SigNoz, so no reason to put extra effort into going for `http/protobuf`
3. A lot of default values are respecified in the config, which makes the config suggested in the guide unnecessary complex.

I can update the guide in signoz-web if you would have my contribution.

LarsSven:
I am a new user of SigNoz who works a lot with Quarkus, and while the Quarkus guide has been of amazing quality, there are some issues with the guide that this MR corrects.

In order of importance:
1. `quarkus.otel.traces.exporter=otlp` does not work. Setting it crashes Quarkus. According to the Quarkus devs, this option has only ever been half-completed and never worked, and instead users should just use the `cdi` exporter, which is also an otlp exporter and is the default exporter https://github.com/quarkusio/quarkus/issues/48570
2. `quarkus.otel.exporter.otlp.protocol=http/protobuf` doesn't seem to work. I'm not sure why, but when I use it, Quarkus seems to be able to send the data just fine to SigNoz without any issues, but it never gets ingested into SigNoz. The gRPC method however works perfectly fine and it is the default method in Quarkus and much better maintained (see https://github.com/quarkusio/quarkus/issues/48572#event-18296258707 for something that was broken for http/protobuf but was not for gRPC as an example). We also have to switch the port from 4318 (http/protobuf otel) to 4317 (gRPC otel).
3. Quarkus includes a BOM when you create a Quarkus application, so there is generally no need to set a version for its extension dependencies (such as `quarkus-opentelemetry`. Removing this version also makes sure the guide doesn't go out of date.
4. There are a fair number of properties in the application.properties that are set to their default value explicitly. I see no reason to respecify them and make the guide more complicated when the default value works just fine. Better to focus on the things that actually need to change.

Resolves https://github.com/SigNoz/signoz/issues/8355

### DIFF
--- a/data/docs/instrumentation/opentelemetry-quarkus.mdx
+++ b/data/docs/instrumentation/opentelemetry-quarkus.mdx
@@ -38,7 +38,7 @@ quarkus extension add opentelemetry
 <TabItem value="Gradle" label="Gradle">
 ```bash
 # Using gradle, add the following to your build.gradle file
-implementation 'io.quarkus:quarkus-opentelemetry:3.17.2'
+implementation 'io.quarkus:quarkus-opentelemetry'
 ```
 </TabItem>
 </Tabs>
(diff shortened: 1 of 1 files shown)


LarsSven:
The troubleshooting part of the guide still contains the old 4318 port, but I don't think it makes much sense to change that port since then the logs would not make sense anymore https://signoz.io/docs/instrumentation/opentelemetry-quarkus/#troubleshooting

Calm-Rock:
Thanks for raising the MR @LarsSven! These changes are helpful.

Just to confirm, the suggested changes have been tested by you, right?

LarsSven:
Yes, I deployed our own Quarkus + SigNoz setup yesterday using this exact setup.

Calm-Rock (approved):

Calm-Rock:
> Yes, I deployed our own Quarkus + SigNoz setup yesterday using this exact setup.

Awesome work @LarsSven! Merging the PR.

Is there anything else you think we should include in our Quarkus documentation? Any suggestions on what can be improved ?

LarsSven:
Some things that could be improved:
- logging and metrics guide via otel (not just tracing)
- include gradle kts in tabs too, not just gradle groovy
- maybe prometheus scraping guide?

PR #1618, opened 2025-06-27
correct reciever to receiver

mznet:
While going through opentelemetry-log doc, I noticed a few typos: “recieve” should be corrected to “receive”. Just in case “recieve” was a valid word, I looked it up on the Internet, but it turns out it’s not correct. While correcting it, I came across a few more “recieve” typos and fixed them.


Calm-Rock (approved):

Calm-Rock:
Thanks for the contribution @mznet! ✌️

Merging the PR.

PR #1715, opened 2025-07-21
Update - opentelemetry grpc vs http

yuvraajsj18:


yuvraajsj18:
Items to verify:

The article claims "15-20% of enterprise networks utilize layer-7 inspection systems that cannot properly process HTTP/2 traffic." This specific statistic could not be verified through current research and appears to be unsourced. While HTTP/2 compatibility issues with enterprise firewalls do exist, no recent studies provide this specific percentage. The claim should either be removed or replaced with a more general statement about potential compatibility challenges without specific percentages.

The article provides very specific performance benchmarks like "4.9x higher throughput at 20ms network latency" and "6.9x higher throughput at 200ms latency" without citing sources. While recent research confirms gRPC generally outperforms HTTP, these exact multipliers are not found in current studies. The 2024 ServiceNow case study shows 500-600K spans/second throughput, but the specific multiplier claims in the article appear to be either outdated or from non-public sources. These should be updated with current, verifiable benchmarks or clearly marked as estimates.

The article claims Protocol Buffers provide "25-40% bandwidth savings" compared to JSON. Current research from 2024-2025 shows Protocol Buffers can be 30-70% smaller than JSON, meaning the article's range is conservative but should be updated to reflect the more recent and broader range of efficiency improvements found in current studies.

The article mentions HTTP/3 with QUIC as delivering specific performance improvements and bridging performance gaps, but presents these as established facts rather than emerging potential. As of 2025, HTTP/3 adoption is still growing and many of the performance claims remain projective rather than proven in production OpenTelemetry deployments. The section should be reframed to clarify that HTTP/3 benefits are expected/potential rather than established in OTLP contexts.

The article includes specific compression ratio claims like "gRPC's stream-aware compression achieves 15-60% ratios" and "HTTP's per-request compression achieves 10-40% ratios" without proper context or sources. These numbers appear to be estimates and should either be sourced or presented as approximate ranges based on typical scenarios rather than definitive measurements.

PR #1724, opened 2025-07-22
Update - open source log management

yuvraajsj18:


yuvraajsj18:
Items to verify:

The claim that the Fluent ecosystem has "over 1.5 billion downloads" is significantly outdated and inaccurate. According to current 2025 data, Fluent Bit alone has surpassed 15 billion downloads globally, making the 1.5 billion figure understated by a factor of 10 or more. This should be updated to reflect that Fluent Bit alone exceeds 15 billion downloads, with the broader Fluent ecosystem having substantially more combined downloads.

The article states that ELK Stack's LogsDB mode provides "2.5x storage reduction," but current information from 2025 indicates that LogsDB mode actually achieves approximately 65% reduction in disk footprint, which is significantly more substantial than the 2.5x figure suggests. The article should be updated to accurately reflect this ~65% storage reduction benefit.

The article doesn't mention OpenSearch 3.0, which was a major release in 2025 that includes significant improvements such as upgrades to Apache Lucene 10, massive vector search speed improvements (up to 9.5x faster), GPU acceleration, and Model Context Protocol (MCP) support. Since this is a 2025 guide, this major release should be acknowledged to provide complete current information.

The article mentions that ELK achieves "220,000 events/second ingestion" but doesn't provide context for this benchmark, and current information suggests that performance varies significantly based on configuration and deployment complexity. The article should clarify that these performance figures are dependent on proper tuning and hardware configuration.

PR #1727, opened 2025-07-23
Update - kubectl logs

yuvraajsj18:


yuvraajsj18:
Items to verify:

The article states that NodeLogQuery feature gate "requires NodeLogQuery feature gate" without specifying its current status. The NodeLogQuery feature gate was introduced as Alpha in Kubernetes 1.27 but has since progressed to Beta status in Kubernetes 1.30. The article should clarify that NodeLogQuery is in Beta (not Alpha) as of Kubernetes 1.30 and remains disabled by default, requiring explicit enabling along with kubelet configuration options.

The RBAC configuration example is incomplete and incorrect. The example shows:
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: log-reader
rules:
- apiGroups: [""]
  resources: ["pods/log"]
  verbs: ["get", "list"]
```
However, the article presents it without the required `rules` field structure and `apiGroups`. The correct format must include the `rules` array with proper `apiGroups` specification. Also, `list` verb is typically not needed for logs access - `get` is sufficient.

The article should mention recent developments in Kubernetes logging that have emerged since early 2025, particularly the Kubernetes History Inspector (KHI) tool announced by Google Cloud in March 2025. This tool provides chronological visualization of cluster logs and represents a significant advancement in Kubernetes log analysis beyond kubectl logs capabilities. Including this would make the article more current and comprehensive.

PR #1729, opened 2025-07-23
Update - kubectl logs

yuvraajsj18:


yuvraajsj18:
Items to verify:

The article includes an incorrect kubectl logs command example using "--since-time='2025-07-23T10:00:00Z'". The kubectl logs command does NOT have a --since-time flag. The correct flag for time-based filtering is --since, which accepts duration values like "1h", "30m", etc., not absolute timestamps. The example should be corrected to use "--since=1h" or similar duration-based syntax instead of the non-existent --since-time flag.

yuvraajsj18:
> Items to verify:
> 
> The article includes an incorrect kubectl logs command example using "--since-time='2025-07-23T10:00:00Z'". The kubectl logs command does NOT have a --since-time flag. The correct flag for time-based filtering is --since, which accepts duration values like "1h", "30m", etc., not absolute timestamps. The example should be corrected to use "--since=1h" or similar duration-based syntax instead of the non-existent --since-time flag.

This works - https://kubernetes.io/docs/reference/kubectl/generated/kubectl_logs/

ankit01-oss (approved):

PR #1737, opened 2025-07-23
Update - docker logs tail

yuvraajsj18:


yuvraajsj18:
Items to verify:

In the logging drivers table, the article states that the journald driver has "✅ Supported" for docker logs support, but this is incorrect. The journald logging driver does NOT support the `docker logs` command. Users must use `journalctl` instead to access logs when using the journald driver. This should be changed to "❌ Use `journalctl`" to match the actual behavior.

In the logging drivers table, for the syslog driver, the article incorrectly states "❌ Use `journalctl`" in the docker logs Support column. This is wrong - syslog driver requires syslog tools, not journalctl. It should be corrected to "❌ Use syslog tools" since logs go to the syslog daemon, not to journald.

The Docker Compose configuration uses `gliderlabs/logspout:v3.2.14` which appears to be an older version. According to current best practices and the project's recommendations, it's better to use `gliderlabs/logspout:latest` or `gliderlabs/logspout:master` instead of pinning to potentially outdated specific version numbers, as the project recommends using these tags for the most up-to-date versions.

ankit01-oss (approved):

PR #1740, opened 2025-07-24
Update - opentelemetry nestjs

yuvraajsj18:


yuvraajsj18:
Items to verify:

The package version for @opentelemetry/sdk-node is incorrect in the installation command. The article shows "npm install --save @opentelemetry/sdk-node@^2.0.0" but the correct version format is 0.x.x (like 0.203.0). OpenTelemetry Node.js SDK uses a 0.x.x versioning scheme, not 2.x.x. The correct installation command should be "npm install --save @opentelemetry/sdk-node@latest" or specify the actual version like "@opentelemetry/sdk-node@^0.203.0".

The NestJS version requirement is slightly misleading. The article states "NestJS 8.0+" as a requirement, but according to the OpenTelemetry NestJS instrumentation package (@opentelemetry/instrumentation-nestjs-core), it supports "@nestjs/core" versions ">=4.0.0 <12". While NestJS 8.0+ is supported, it's not the minimum requirement - the minimum is actually NestJS 4.0. The article should clarify that NestJS 4.0+ is supported, with 8.0+ being recommended for better compatibility.

The Node.js version requirements context could be more precise. While the article correctly states "Node.js 18.19.0+ or Node.js 20.6.0+" for OpenTelemetry SDK 2.0, it should also mention that NestJS 11+ requires Node.js 20 or higher, which affects the overall compatibility matrix when using both technologies together.

ankit01-oss (approved):

PR #1745, opened 2025-07-24
Update - open source apm tools

yuvraajsj18:


yuvraajsj18:
Items to verify:

The article claims "96% enterprise adoption" for OpenTelemetry, but according to 2025 data from multiple sources, only about 48.5% of organizations are currently using OpenTelemetry, with another 25.3% planning to implement it soon. The 96% figure appears to be significantly inflated and should be corrected to reflect the actual ~48.5% current adoption rate.

The article claims "43% increase in GitHub commits" for OpenTelemetry but no supporting data was found for this specific statistic. Instead, sources mention an "18% increase in developer participation and a 22% rise in company involvement year over year." The 43% GitHub commits increase claim should either be removed or replaced with verified statistics like the 18% increase in developer participation.

The article states SigNoz has "17k+ GitHub stars" but current data from mid-2025 shows approximately 19.1k stars. While this is not technically inaccurate since 19.1k is indeed "17k+", it would be more precise to update this to "19k+" to reflect the current higher star count.

Several GitHub star counts could not be verified from current sources (Jaeger: claimed 20k+, Prometheus: claimed 55k+, Grafana: claimed 60k+, Zipkin: claimed 17k+). While these may be accurate, they should be verified by checking the repositories directly to ensure the information is current for 2025.

The performance benchmark ranges provided (like "Jaeger: 15-30% throughput impact, Zipkin: 10-25% throughput impact") are presented as if they are from 2025 data, but limited benchmarking data was found to support these specific ranges. Only Zipkin showed ~16% CPU overhead with 100% sampling in available sources. These performance claims should either be supported with current sources or marked as approximate/estimated ranges rather than definitive 2025 benchmarks.

PR #1768, opened 2025-07-30
Update - nginx logging

yuvraajsj18:


yuvraajsj18:
Items to verify:

The article claims NGINX powers "over 35% of all websites globally", but recent data shows varying market share figures: 38.6% (Monitor.Us), 33.8% (W3Techs), and 21.2% (Netcraft) depending on measurement methodology. While "over 35%" could be considered accurate based on some sources, it represents the higher end of the range. The claim should be updated to reflect the variation in measurements and provide a more current, specific figure with source attribution.

The JSON logging format example includes "$request_id" as a standard NGINX variable, but $request_id is not available in the core NGINX installation. It requires the ngx_http_request_id_module or NGINX Plus. Using this variable without the proper module will cause configuration errors. The example should either remove this variable or include a note explaining that it requires additional modules to be enabled.

The logrotate configuration uses "invoke-rc.d nginx rotate" in the postrotate section, but this command is not universally reliable and may fail in some environments. The standard and recommended approach is to use "kill -USR1 `cat /run/nginx.pid`" or similar signal-based commands to instruct NGINX to reopen log files. The configuration should be updated to use the more reliable USR1 signal method.

The OpenTelemetry timestamp parsing layout '%Y-%m-%dT%H:%M:%S%z' may not work correctly with all ISO 8601 timestamps. Specifically, if NGINX outputs timezone offsets with colons (e.g., +02:00), this layout will fail because the OpenTelemetry Collector's filelog receiver (using Go's time parsing) doesn't support colons in timezone offsets. The article should mention this limitation and provide guidance on timestamp format compatibility.

PR #1818, opened 2025-08-12
feat: PYTHON DOCS - otel-collector inside docker 

anukulpandey:


anukulpandey:
will be using https://github.com/SigNoz/signoz-web/pull/1889 for all python lib's docker instructions

PR #1845, opened 2025-08-21
chore: added dashboard migration note

Calm-Rock:


makeavish:
Add comment in individual page as well

Calm-Rock:
> Add comment in individual page as well

We don't have individual pages for these dashboards right now. We just redirect to GitHub. Will add the note when we have these pages.

makeavish (approved):

PR #1891, opened 2025-09-01
single doc for python logging

### LINKED ISSUE #1421: Improve docs on logs collection (Python)
Doc on ideal way of sending logs from Python application.

Currently our helm chart by default collects logs(along with metadata like k8s namespace) but we don't mention that we need to parse most attributes with pipelines.

Our python direct logs collection doc doesn't even have mention of logs correlation.

Calm-Rock:
Fixes https://github.com/SigNoz/signoz-web/issues/1421


data/docs/logs-management/send-logs/python-logs.mdx
```diff
@@ -0,0 +1,67 @@
...
+title: Sending Python Logs
+id: python-logs
+hide_table_of_contents: true
+---
+
+You can send Python logs to SigNoz using Auto instrumentation which is the fast, no code change way to send logs to SigNoz.
+
+## Prerequisites
+
+- A [auto-instrumented](https://signoz.io/docs/instrumentation/opentelemetry-python/#send-traces-directly-to-signoz-cloud) python application with logging implemented
```
makeavish: This prerequisite is not valid when user is on k8s and don't want to instrument app

PR #1895, opened 2025-09-02
fix: add missing pipelines value to httpcheck example

Rick-VA:
I was following this guide but noticed that the pipelines value was missing from the examples.


Calm-Rock:
Hey, @Rick-VA !
Thank you for the PR and for contributing to the improvement of SigNoz docs. 

Have approved the PR.

Calm-Rock (approved):

PR #1955, opened 2025-09-11
fix: issue with 8080 port already bind issue in the signoz in ECS setup

ravimahey:
By default, SigNoz uses port 8080.
In this setup, ZooKeeper also runs its admin server on port 8080. Because the wrong environment variable (`ZOOKEEPER_ADMIN_SERVER_PORT_NUMBER`) was used, ZooKeeper did not correctly configure its admin server port, which led to a direct conflict with SigNoz 8080 port.

To fix this, update the environment variable to the correct one:

> ZOO_ADMIN_SERVER_PORT_NUMBER

This resolves the port binding conflict on 8080.

WhatsApp Image 2025-09-11 at 23 19 29_91099d9b


Nageshbansal (approved):

Nageshbansal:
Related: https://github.com/SigNoz/platform-pod/issues/568

PR #2014, opened 2025-09-29
Update Python Logging Best Practices Article

yuvraajsj18:


Elizabeth-Mathew1:
> Log Levels: You can categorize messages by severity (e.g., DEBUG, INFO, WARNING, ERROR), allowing you to filter out noise.1

Can we remove the extra 1?

> Configurability: You can easily change the logging output: sending it to the console, a file, or a network service, without changing your application code.

The continous ":" hinders flow of reading and looks repetetive. Can we use a hyphen instead?

Do you think along with log levels we should also maybe write very briefly ablut log numbers also? [only if needed]

>  A ContextVar holds
can we make it contextvar?

Rest LGTM!

Elizabeth-Mathew1 (approved):

PR #2082, opened 2025-10-16
adding pydantic ai docs

gkarthi-signoz:


pranay01:
@gkarthi-signoz If someone wants to monitor all of metrics, traces and logs will they have to add the steps you have mentioned in each section? Will just auto-instrumentation not work?

@makeavish do you have any understanding if just auto instrumentation will work?

makeavish:
> @gkarthi-signoz If someone wants to monitor all of metrics, traces and logs will they have to add the steps you have mentioned in each section? Will just auto-instrumentation not work?
> 
> @makeavish do you have any understanding if just auto instrumentation will work?

Auto instrumentation should work. Eg: https://www.braintrust.dev/docs/integrations/pydantic-ai

pranay01:
@makeavish the braintrust docs you gave still needs TracerProvider() to be created.  So, this is still not complete auto instrumentation, right? I was thinking if something like https://signoz.io/docs/instrumentation/opentelemetry-python/#send-traces-to-signoz-cloud would work

makeavish:
> @makeavish the braintrust docs you gave still needs TracerProvider() to be created. So, this is still not complete auto instrumentation, right? I was thinking if something like https://signoz.io/docs/instrumentation/opentelemetry-python/#send-traces-to-signoz-cloud would work

It should be possible to just use opentelemetry-instrument instead of TracerProvider() as it also enables global tracer which is needed by Agent.instrument_all(). Official docs reference - https://pypi.org/project/opentelemetry-instrumentation/?utm_source=chatgpt.com#:~:text=This%20will%20also%20configure%20a%20global%20tracer%20and%20tracing%20exporter%20as%20well%20as%20a%20meter%20and%20meter%20exporter

I suggest @gkarthi-signoz verifies this and use auto instrumentation where ever possible

pranay01 (approved):

PR #2090, opened 2025-10-19
llm docs seo edits

gkarthi-signoz:


pranay01:
@gkarthi-signoz I see you have replaced monitoring with observability in most of the cases. Is it informed by the search volume in each case or did you just go with changing it to observability for all cases

gkarthi-signoz:
> @gkarthi-signoz I see you have replaced monitoring with observability in most of the cases. Is it informed by the search volume in each case or did you just go with changing it to observability for all cases

i used ahref to see which keyword was higher for each docs. Some docs I left unchanged with "monitoring" because search volume is higher with that. Only few I replaced with observability

pranay01 (approved):

pranay01:
> > @gkarthi-signoz I see you have replaced monitoring with observability in most of the cases. Is it informed by the search volume in each case or did you just go with changing it to observability for all cases
> 
> i used ahref to see which keyword was higher for each docs. Some docs I left unchanged with "monitoring" because search volume is higher with that. Only few I replaced with observability

ok merged

data/docs/llamaindex-observability.mdx
```diff
@@ -1,14 +1,14 @@
 ---
 date: 2025-08-1
-id: llamaindex-monitoring-with-opentelemetry
-tags : [SigNoz Cloud]
-title: Monitoring LlamaIndex Telemetry with SigNoz
-description: Enable telemetry in your LlamaIndex application with OpenTelemetry and send traces, logs, and metrics to SigNoz
+id: llamaindex-observability-with-opentelemetry
+tags : [SigNoz Cloud, LlamaIndex, Observability, Monitoring, OpenTelemetry]
```
makeavish, line 4: These tags don't exist - https://github.com/SigNoz/signoz.io/blob/78c0e2ea249cdb471c771cf52abdfcee2b47ed20/constants/tagDefinitions.ts#L4
Please remove non existent tags

Also these docs should work for Self hosted as well right? Remove SigNoz Cloud tab and where ever we are mentioning ingestion URL just ask them to use: `http://<otel-collector-host>:4318` or `:4317` (OTLP/HTTP vs OTLP/gRPC).

PR #2093, opened 2025-10-20
feat: add docs redirect guard and pre-commit hook for validation

makeavish:


scripts/check-doc-redirects.js
```diff
@@ -0,0 +1,182 @@
...
+  const withoutPrefix = docPath.replace(/^data\/docs\//, '').replace(/\.mdx$/, '')
+  const segments = withoutPrefix.split('/').filter(Boolean)
+  if (segments[segments.length - 1] === 'index') {
+    segments.pop()
+  }
+  const route = `/docs${segments.length ? `/${segments.join('/')}` : ''}/`
+  return route.replace(/\/+/g, '/')
+}
+
+function readRedirects() {
```
manika-signoz, line 28: can this method also handle circular redirects? (A -> B -> C -> A ?) is it useful to add this?

.husky/pre-commit
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
+
+set -e
+
+echo 'husky (pre-commit): running lint-staged'
+yarn lint-staged
+
+STAGED_FILES=$(git diff --cached --name-only)
+
+if printf '%s\n' "$STAGED_FILES" | grep -E '^(data/docs/.*\.mdx|next\.config\.js|scripts/check-doc-redirects\.js)$' >/dev/null; then
```
manika-signoz, line 10: can we make this generic if possible, the pre commit running for every data/ folder change instead of just docs?

makeavish: Intentionally reduced scope to docs. Blog team can expand this if needed.

scripts/check-doc-redirects.js
makeavish: No it doesn't. Can be implemented but very uncommon case. I will like to keep script simple for now and skip this.

.husky/pre-commit
manika-signoz: okay

manika-signoz (approved):

PR #2153, opened 2025-11-03
add test commit

manika-signoz:
1. Added <GetStartedSigNoz /> to are-there-any-hidden-costs-associated-with-using-signoz (faqs)


manika-signoz:
not needed anymore

PR #2197, opened 2025-11-12
feat: add descs to case-study and remove slug from frontmatter in faqs and case-study

manika-signoz:


data/case-study/blip.mdx
```diff
@@ -1,23 +1,18 @@
 ---
 title: How Blip uses SigNoz to improve their issue resolution time by 14x
-slug: blipbillboards
+description: How Blip uses SigNoz to improve their issue resolution time by 14x
```
ahrefabhi, line 3: why did we remove slug, is it not required or we are generating it in strapi?

ahrefabhi (approved):

data/case-study/blip.mdx
ahrefabhi: You've removed slug from case-study also, but the PR title says you've removed it from faqs.

manika-signoz: the field is not present in strapi and needs to be removed from all documents, apologies for the miss, it has been removed from both faqs and case-study
the sync workflow would fail in case the frontmatter contains a field that is not present in the strapi collection type, the path the document needs to be rendered on is mapped to the "path" field, it is picked from the filename / the location the file is present in, that's why it is safe and important to remove slug from frontmatter

yuvraajsj18 (dismissed):

ahrefabhi (dismissed):

ahrefabhi (approved):

PR #2202, opened 2025-11-12
fix(css): update doc layout to fix width dynamic changes on content changes

makeavish:
Change video with before after behaviour - https://www.loom.com/share/6c25c895125e4cfeabb072ff96e985c2

### DIFF
--- a/css/doc.css
+++ b/css/doc.css
@@ -4,12 +4,15 @@
   height: 100%;
   gap: 16px;
   max-width: 1504px;
+  width: 100%;
+  align-items: flex-start;
 }
 
 .doc-toc {
   width: 320px;
   min-width: 320px;
   max-width: 320px;
+  flex: 0 0 320px;
   box-sizing: border-box;
   height: calc(100vh - 156px);
   overflow-y: auto;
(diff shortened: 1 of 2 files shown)


css/doc.css
```diff
@@ -36,11 +39,13 @@
...
   min-width: 320px;
   max-width: 320px;
+  flex: 0 0 320px;
   box-sizing: border-box;
 }
 
 .doc-content {
-  width: calc(100% - 320px);
+  flex: 1 1 auto;
+  min-width: 0;
```
manika-signoz, line 48: this might impact documentation on onboarding flows, have we validated this?

makeavish: Nice catch, let me check that

css/global.css
```diff
@@ -1,6 +1,13 @@
 html {
   scroll-padding-top: 60px; /* Adjust this value to match the height of your sticky header */
   scroll-behavior: smooth;
+  overflow-y: scroll; /* keep scrollbar gutter reserved even when content is short */
+  scrollbar-gutter: stable both-edges;
```
manika-signoz, line 5: https://github.com/user-attachments/assets/13c7803e-feec-4b90-8e77-aaa18917c8f3

this is causing the root page to have a small empty space on the right which is not present on signoz.io currently (please refer to the video)

PR #2203, opened 2025-11-13
chore: disable lint staged

makeavish:

PR #2204, opened 2025-11-13
fix: fix linter config and allow override

manika-signoz:


manika-signoz:
closing in favour of https://github.com/SigNoz/signoz.io/pull/2203

PR #2208, opened 2025-11-14
Skip Chatbase bubble when onboarding source param

yuvraajsj18:
## Summary
- skip loading the Chatbase embed when the page URL includes `source=onboarding`
- prevent the floating bubble from appearing during onboarding flows

### DIFF
--- a/components/Chatbase/ChatbaseClient.tsx
+++ b/components/Chatbase/ChatbaseClient.tsx
@@ -2,6 +2,8 @@
 
 import React, { useEffect, useRef, useState } from 'react'
 import Script from 'next/script'
+import { useSearchParams } from 'next/navigation'
+import { ONBOARDING_SOURCE } from '@/constants/globals'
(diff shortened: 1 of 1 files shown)


components/Chatbase/ChatbaseClient.tsx
```diff
@@ -30,6 +30,14 @@ export default function ChatbaseClient({
...
     }
 
+    const hasOnboardingSourceParam =
+      new URLSearchParams(window.location.search).get('source') === 'onboarding'
+
+    if (hasOnboardingSourceParam) {
+      console.log('Skipping Chatbase initialization due to onboarding source param')
+      return
+    }
+
```
manika-signoz, line 41: please avoid using window.location.search
use useSearchParams() hook from next and .get() on returned searchParams object and use the constant defined in globals to match against 
import { ONBOARDING_SOURCE } from '@/constants/globals'

manika-signoz (approved):

PR #2233, opened 2025-11-20
updating field attributes

revmag:
1) Updated documentation to state that ambiguity now results in a warning instead of an error.
2) Added notes explaining how conflicts between attributes, resources, and other field types are resolved in different contexts.


nityanandagohain (commented):
The documentation changes looks good to me, not sure about the impact of change in app/tag-data  @makeavish

makeavish (approved):

PR #2237, opened 2025-11-20
chore: update node version for local dev env

manika-signoz:
Closes https://github.com/SigNoz/engineering-pod/issues/3154


app/tag-data.json
```diff
@@ -1,146 +1,81 @@
 {
-  "opentelemetry": 162,
-  "signoz": 43,
+  "product": 39,
+  "opentelemetry": 115,
+  "signoz": 40,
   "community": 7,
-  "prometheus": 75,
+  "prometheus": 9,
   "distributed-tracing": 19,
```
ahrefabhi, line 7: what's the use of this file @manika-signoz ?

manika-signoz: this file gets regenerated on build and powers https://signoz.io/tags/ page here
it doesn't matter if it is committed or not, since it is always regenerated on build, we have this in gitignore but must have gotten added to the PR due to some git cache, can be ignored. i'll make sure to discard these changes from next time before committing

ahrefabhi (approved):

PR #2277, opened 2025-11-28
[OpenAPI] Fix response schemas

SferaDev:
Update the OpenAPI specification to standardize API responses and error handling:

- Add missing response schemas
- Fix incorrect response schemas (`FieldsResponse` and `DashboardResponse`)
- Added new `ApiSuccessResponse` and `ApiErrorResponse` schemas to wrap all successful and error responses

The biggest change is that several endpoints return an envelope `{ success: true; data: Data } | { success: false; error: Error }` while the spec documents the response to be directly `Data`.

### DIFF
--- a/public/openAPISpec/api.yaml
+++ b/public/openAPISpec/api.yaml
@@ -65,8 +65,24 @@ paths:
             type: string
           example: '<TOKEN>'
       responses:
+        '200':
+          description: Field updated successfully
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Field'
         '400':
           description: Invalid input
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ApiErrorResponse'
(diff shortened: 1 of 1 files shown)


SferaDev:
@therealpandey I don't see any changes in https://signoz.io/openAPISpec/api.yaml, also although https://github.com/SigNoz/signoz/blob/8771919de632baafb2aaffcf91f0bc392b33645a/docs/api/openapi.yml correctly represents the response wrapper it's missing a lot of endpoints (almost all of the core functionality of signoz)

therealpandey:
Hi @SferaDev,
Thanks for the input.

We'll be pushing more endpoints there. The idea is to keep the spec generation as close to code as possible to ensure they dont go out of sync.

SferaDev:
Sounds good, could we get query_range soonish, that's the most important endpoint 😅

PR #2336, opened 2025-12-17
Add doc explaining the misconceptions about droppping metric labels

srikanthccv:
It has been observed that users attempt to drop the labels in an attempt to reduce the cardinality and bill but lable deletion doesn't help and in fact results in incorrect results. This guide explains it with examples.


srikanthccv:
@abhishekhugetech @nikhilmantri0902 FYI.

package.json
```diff
@@ -105,5 +105,6 @@
     "*.+(js|jsx|ts|tsx|json|css)": [
       "prettier --write"
     ]
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
manika-signoz: why was this added?

srikanthccv: I think my current yarn@1.22.22 add this automatically. Didn't notice; Will remove.

data/docs/metrics-management/dropping-metric-labels.mdx
```diff
@@ -0,0 +1,184 @@
...
+
+### 2. Incorrect Aggregations
+
+When you query and aggregate these metrics, the math becomes meaningless:
+
+Without the `k8s.pod.*` label, you can't properly distinguish between pods. The aggregation may double-count, miss data, or produce values that don't reflect reality.
+
+## The Correct Approach: Aggregation
+
+If you genuinely don't need granularity at a certain attribute level, the correct approach is to **aggregate** the values mathematically, not just drop the label. Use the `metricstransformprocessor`:
```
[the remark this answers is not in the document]
srikanthccv: not true, it's not deprecated. they both have their own purpose.

package.json
srikanthccv: updated.

srikanthccv:
@claude, my brother, living in the far away in the cloud. addressed comments. PTAL.

manika-signoz (approved):

makeavish (approved):

therealpandey (approved):

srikanthccv:
@makeavish, last commit has some inaccuracy, `k8s.pod.uid` in scenario 1 is an identifying label (the section title says removing non-identifying label). Both the scenario remove the identifying labels (one can freely drop non-identifying label without any worry). I added two examples to convey,

1. it's generally recommended to not put IDs in metrics, so people remove ID, while it's not recommended, the result may not necessarily contain the duplicate time series because the existence of other attribute may save them from inaccuracies (k8s.pod.name, which is not guaranteed to unique, but may appear to be unique for deployments, I had one customer who argued this and had to explain him). 
2. in second, removed both pod name and id, to show that it leads to duplicate series resulting in inaccuracies for sure.


Please adjust the content accordingly.

therealpandey (approved):

PR #2400, opened 2026-01-02
test: sync for otel folders

manika-signoz:


manika-signoz:
closing since not needed anymore

PR #2555, opened 2026-01-19
docs: enhance troubleshooting for Gunicorn/Uvicorn worker spans

adionit7:
## Pull Request

---

Users deploying FastAPI/Python apps with Gunicorn or Uvicorn multiple workers often experience missing spans in SigNoz. The existing troubleshooting section mentions this briefly but doesn't explain the root cause or provide a concrete solution.

This PR enhances the troubleshooting section with:
- Clear explanation of why spans drop (forking issue with OpenTelemetry SDK)
- Complete `gunicorn_config.py` example with `post_fork` hook
- Link to the FastAPI production example for a full working reference
- Link to official OpenTelemetry multiprocessing docs

This helps users understand the problem and implement the fix themselves, reducing support load.

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only
- [x] 📚 Documentation

---

N/A - Documentation enhancement.

---

- Tests added/updated: N/A (documentation)
- Manual verification: 
  - Verified markdown syntax
  - Checked code example syntax
  - Confirmed links are valid
  - Reviewed content for accuracy
- Edge cases covered: N/A

---

- Blast radius: None - documentation only
- Potential regressions: None
- Rollback plan: Simple revert if needed

---

| Field | Value |
|------|-------|
| Deployment Type | OSS |
| Change Type | Maintenance |
| Description | Enhanced troubleshooting section for Gunicorn/Uvicorn worker span issues with code example and links |

---

### 📋 Checklist
- [x] Tests added or explicitly not required (documentation)
- [x] Manually tested (verified formatting and links)
- [x] Breaking changes documented (N/A)
- [x] Backward compatibility considered (N/A)

---

## 👀 Notes for Reviewers

This addresses the issue mentioned in the tweet thread where users struggle with Gunicorn workers dropping spans. The existing troubleshooting section was too brief - this adds the detailed explanation and working code example that users need.

The `gunicorn_config.py` example matches the pattern used in the FastAPI production example we added to the examples repo, ensuring consistency across documentation.

---

### DIFF
--- a/data/docs/instrumentation/opentelemetry-python.mdx
+++ b/data/docs/instrumentation/opentelemetry-python.mdx
@@ -578,43 +578,20 @@ If you see JSON span output in your terminal but traces don't appear in SigNoz, 
 
 Application servers that spawn multiple worker processes require special handling because the OpenTelemetry SDK isn't fork-safe.
 
-**The problem:** When Gunicorn or Uvicorn forks worker processes, the OpenTelemetry SDK state (spans, exporters, etc.) initialized in the parent process isn't properly inherited by worker processes. This means spans created in workers often don't get exported.
+**The problem:** When Gunicorn or Uvicorn forks worker processes, the OpenTelemetry SDK state (TracerProvider, exporters, span processors) initialized in the parent process isn't properly inherited by workers. The background threads that export spans don't survive the fork, so spans created in workers never get sent.
 
-**Solutions:**
+**Quick fixes:**
 
 - **Uvicorn** with `--workers` flag is not supported. Use Gunicorn with Uvicorn workers instead:
   ```bash
   opentelemetry-instrument gunicorn -k uvicorn.workers.UvicornWorker main:app
   ```
 
-- **Gunicorn with multiple workers:** For production deployments, you need to reinitialize OpenTelemetry in each worker after forking. Create a `gunicorn_config.py`:
-
-  ```python
-  import os
-
-  def post_fork(server, worker):
-      """Reinitialize OpenTelemetry in each worker after fork"""
-      if os.getenv("OTEL_SERVICE_NAME"):
-          os.system("opentelemetry-bootstrap --action=install")
-
-  bind = "0.0.0.0:8000"
-  workers = 4
-  worker_class = "uvicorn.workers.UvicornWorker"
-  preload_app = False  # Don't preload with OpenTelemetry
-  ```
-
-  Then run:
-  ```bash
-  opentelemetry-instrument gunicorn main:app -c gunicorn_config.py
-  ```
-
-  A complete working example with Docker and production configuration is available in the [FastAPI production demo](https://github.com/adionit7/examples/tree/feat/fastapi-production-demo/fastapi/fastapi-production-demo).
-
-- **Hypercorn/Unicorn** are not supported due to fork-safety issues. See the [Hypercorn/Unicorn tab](#framework-instrumentation) for details and workarounds.
+- **Gunicorn with multiple workers:** You need to reinitialize the TracerProvider in each worker after forking using a `post_fork` hook. See [Running with Gunicorn or uWSGI](#running-with-gunicorn-or-uwsgi) for setup details and the <a href="https://github.com/open-telemetry/opentelemetry-python/tree/main/docs/examples/fork-process-model" target="_blank" rel="noopener noreferrer nofollow">official post_fork example</a> for the correct implementation.
 
 - **Gunicorn with `--preload`** or **uWSGI** need extra config. See [Running with Gunicorn or uWSGI](#running-with-gunicorn-or-uwsgi).
 
-For more details, see the [OpenTelemetry Python multiprocessing documentation](https://opentelemetry-python.readthedocs.io/en/latest/instrumentation/runtime.html#multiprocessing).
+- **Hypercorn/Unicorn** are not supported due to fork-safety issues. See the [Hypercorn/Unicorn tab](#framework-instrumentation) for details and workarounds.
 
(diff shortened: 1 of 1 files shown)


shubham-signoz (commented):
Hey, thanks for taking the time to improve the docs here!

A couple of things though - the post_fork example won't quite work as written. `opentelemetry-bootstrap -a install` is actually a [package installer](https://opentelemetry.io/docs/languages/python/automatic/) so calling it at runtime on every worker fork isn't what we want or the right way to do it. (With 4 workers you'd have 4 pip install attempts on startup, and more each time workers recycle. In most production setups this just fails (read-only containers, no network, slim images without pip). And even if it ran, it's not a reliable fix since it might work for some setups but not others - the packages are already there, it's the SDK state (TracerProvider, exporters, processors) that didn't survive the fork).

The correct approach is reinitialising the TracerProvider/MeterProvider in each worker.We do already cover this in the [#running-with-gunicorn-or-uwsgi ](https://signoz.io/docs/instrumentation/opentelemetry-python/#running-with-gunicorn-or-uwsgi)section with links to the official post_fork examples.

That said - I'm curious what prompted this. Was something confusing about the existing pre-fork/post-fork explanation? Would love to hear what tripped you up so we can fix it in the right place.

adionit7:
> shubham-signoz

Thanks for the catch, Shubham! You are right.. running `opentelemetry-bootstrap` at runtime is an anti-pattern for production.

What tripped me up: I was specifically looking for a 'Troubleshooting' section for missing spans and completely missed the `#running-with-gunicorn` link further down. Since the auto-instrumentation docs imply it 'just works', I didn't initially think to look for manual post_fork hooks.

Updates pushed:
- Removed the runtime bootstrap snippet.
- Added a clear pointer/admonition to the existing `#running-with-gunicorn` section so others find it faster.

(Note: I also updated my PR in the examples repo to reflect the correct post_fork approach).

Ready for re-review!

shubham-signoz:
> > shubham-signoz
> 
> Thanks for the catch, Shubham! You are right.. running `opentelemetry-bootstrap` at runtime is an anti-pattern for production.
> 
> What tripped me up: I was specifically looking for a 'Troubleshooting' section for missing spans and completely missed the `#running-with-gunicorn` link further down. Since the auto-instrumentation docs imply it 'just works', I didn't initially think to look for manual post_fork hooks.
> 
> Updates pushed:
> 
> * Removed the runtime bootstrap snippet.
> * Added a clear pointer/admonition to the existing `#running-with-gunicorn` section so others find it faster.
> 
> (Note: I also updated my PR in the examples repo to reflect the correct post_fork approach).
> 
> Ready for re-review!

@adionit7 I think the existing troubleshooting already has a reference for the gunicorn preload. I don't think we need any more additional information except on this line. Do you think maybe we need to make this line more explicit about the issue?
https://github.com/SigNoz/signoz.io/blob/8ef9ed1273c5954926a1a727905127314227b69f/data/docs/instrumentation/opentelemetry-python.mdx?plain=1#L586

adionit7:
@shubham-signoz That makes total sense! I agree.. keeping it concise is better.

The current line (need extra config) is a bit passive, which is why I scrolled past it. I'll update it to explicitly mention 'missing spans' and 'forking issues'. That way, anyone debugging dropped traces will immediately know this is the relevant section.

Updating the PR now to just modify this line.

data/docs/instrumentation/opentelemetry-python.mdx
```diff
@@ -519,9 +519,9 @@ Choose your framework below to see the specific run command. The setup steps abo
 
 </ToggleHeading>
 
-**Gunicorn** works out of the box. No extra setup needed unless you use the `--preload` flag. If you do use `--preload`, see the <a href="https://github.com/open-telemetry/opentelemetry-python/tree/main/docs/examples/fork-process-model" target="_blank" rel="noopener noreferrer nofollow">post_fork hook example</a>.
+**Gunicorn** works out of the box—unless you use `--preload` or multiple workers, which can cause missing spans due to forking issues. See the <a href="https://github.com/open-telemetry/opentelemetry-python/tree/main/docs/examples/fork-process-model" target="_blank" rel="noopener noreferrer nofollow">post_fork hook example</a> for the fix.
```
crazyuploader, line 522: The fork-safety issue is specific to `--preload`: the master process starts the `BatchSpanProcessor` background thread before forking, and child processes inherit the locked state, causing deadlocks. Without `--preload`, each worker initializes the processor independently after the fork — no issues.

Remove "or multiple workers". The original scope (`--preload` only) was correct.

Source: https://github.com/open-telemetry/opentelemetry-python/tree/main/docs/examples/fork-process-model

data/docs/instrumentation/opentelemetry-python.mdx
```diff
@@ -574,17 +574,6 @@ OTEL_TRACES_EXPORTER=console opentelemetry-instrument <your_run_command>
 
 If you see JSON span output in your terminal but traces don't appear in SigNoz, the issue is with export configuration (endpoint, auth, or network). If no output appears, the instrumentation isn't capturing your requests.
 
-### Why do multi-worker servers (Uvicorn, Gunicorn) drop spans?
```
crazyuploader, line 577: Please restore the section, or move the `uvicorn --workers` note into "Running with Gunicorn or uWSGI" before merging.

**The uWSGI change is correct, keep it.**

crazyuploader (changes requested):

PR #2611, opened 2026-01-30
Fix typo in SMTP TLS insecure skip verify variable

bracoTuxbr:
After a long time debugging to get email sending working, I realized the DOC was wrong.

Another link in the DOC talks about underscores, and that's when I discovered I had to change them. Then it worked and stopped returning the message:

failed to send STARTTLS command

The correct code is: SIGNOZ_EMAILING_SMTP_TLS_INSECURE__SKIP__VERIFY=true
instead of
SIGNOZ_EMAILING_SMTP_TLS_INSECURE_SKIP_VERIFY=true


therealpandey (approved):

makeavish:
@bracoTuxbr Thanks for the PR!

PR #2734, opened 2026-02-16
feat: comparisons and guides migration

manika-signoz:


manika-signoz:
concluded, not needed anymore

PR #2882, opened 2026-03-12
feat: why otel doc

shubham-signoz:


crazyuploader (commented):
Add this doc's link in other relevant docs, otherwise looks good to me

shubham-signoz:
@crazyuploader I will do that in different PR .

crazyuploader (approved):

data/docs/overview/what-is-opentelemetry-and-why-it-matters.mdx
```diff
@@ -0,0 +1,89 @@
...
+
+Before OpenTelemetry, each vendor shipped its own agent, SDK, and data model. Pick a monitoring tool, and that choice follows you into every service, every deploy pipeline, every runbook. Most teams just stay put, even when the tool no longer fits.
+
+OpenTelemetry breaks that. Instrument once against a vendor-neutral API. Change backends by updating an exporter endpoint. Dashboards still need rebuilding, but touching every service to swap instrumentation goes away.
+
+OpenTracing and OpenCensus tried this separately. Neither got broad adoption. OpenTelemetry unified both with enough industry backing to sustain it.
+
+## What you get in practice
+
+**Instrument once, send anywhere.** Where your data goes (SigNoz, Jaeger, a cloud provider, all three at once) is a Collector config change. No code rewrite. Want to evaluate a new backend? Add a second exporter and compare side by side.
```
revmag: Content should feel less like marketing.
imo this will mainly be read by AI agents( when people search why is this standard coming, why do we even shift), and then the user. So have to rethink from that perspective.

revmag: Something along the lines of :

If you adopt OpenTelemetry, you are usually choosing to depend less on one vendor’s SDK. If you want to test a different backend, run two backends in parallel, or migrate later, you usually do not have to re-instrument every service. In many cases, the change happens in your export or Collector configuration instead. 


OpenTelemetry gives you a shared format and common attribute names across services and languages. That makes traces easier to follow and reduces the usual naming drift between teams.

Using the Collector lets you move some decisions out of application code: sampling, filtering, redaction, batching, and routing can often be handled centrally. That is useful for larger teams, especially when platform and application responsibilities are separate.

revmag (approved):

therealpandey (approved):

therealpandey (approved):

PR #2934, opened 2026-03-20
fix: shuffle quiz options, show scores as X/3

Elizabeth-Mathew1:
## Summary
- Shuffle answer options per question so the best answer isn't always option A (uses a deterministic seeded shuffle)
- Display skill breakdown scores as X/3 instead of X/10 for clarity

## Test plan
- [ ] Complete the quiz and verify options appear in different order per question
- [ ] Verify the same question always shows the same shuffled order (deterministic)
- [ ] Check results page skill breakdown shows scores as X/3 format
- [ ] Verify scoring still works correctly regardless of option position

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/sre-skill-decay-index/components/QuizSection.tsx
+++ b/app/sre-skill-decay-index/components/QuizSection.tsx
@@ -1,5 +1,6 @@
+import { useMemo } from 'react'
 import Image from 'next/image'
-import { Category, QuizQuestion } from '../types'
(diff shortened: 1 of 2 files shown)


ankit01-oss (approved):

YounixM (approved):
Blanket approval

ankit01-oss (approved):

PR #2985, opened 2026-03-31
fix: resolve multiple H1 tags across 32 pages

yuvraajsj18:
Notion: https://www.notion.so/334fcc6bcd19818da947cd3bf20862a6


yuvraajsj18:
Re: bumped `date` fields in docs files

The date updates were unavoidable. The repo's `check-docs-metadata` hook (also enforced in CI via `docs-metadata-guard.yml`) requires that any staged docs file has a `date` within 7 days of today. Since modifying heading levels in these files means staging them, the hook rejects their original dates (some as old as `2024-11-03`).

Reverting the dates would cause CI to fail. The `description` fields were also added at the same time since those files were missing them — those additions are intentional and accurate.

makeavish (approved):

PR #3006, opened 2026-04-01
fix: add missing and resolve duplicate meta descriptions across docs

yuvraajsj18:
https://www.notion.so/signoz/299fcc6bcd19806388eccb39e69e6d1e

### DIFF
--- a/data/docs/alerts-management/apdex-alerts.mdx
+++ b/data/docs/alerts-management/apdex-alerts.mdx
@@ -1,8 +1,8 @@
 ---
-date: 2025-08-05
+date: 2026-04-01
 id: apdex-alerts
-
 title: Creating Alerts based on Apdex Score
+description: Create alerts based on Apdex scores in SigNoz to monitor application performance using satisfaction thresholds and PromQL queries.
 ---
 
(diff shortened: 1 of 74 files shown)


yuvraajsj18:
Thanks for the review!

**P2.1 — Stray blank lines:** Fixed in the latest commit (209b4db8).

**P2.2 — Date bumping:** Intentional and required. The pre-commit hook runs `scripts/check-docs-metadata.js` which validates that `date` must be within 7 days of today. Files with dates from 2024 would fail that check whenever they are modified in future PRs. Updating the date on every touched file keeps them compliant with the existing gate.

**P3 — `doc_type` missing:** Pre-existing tech debt on these 68 files, noted. Out of scope for this PR (SEO metadata fix only) — can be tracked separately.

**Build OOM:** The Vercel OOM failure is a pre-existing infrastructure issue unrelated to these changes (all edits are frontmatter text, no new pages or assets). The build log itself confirms it: `.git/objects/pack` alone is 1.2 GB. Suggest enabling Enhanced Builds or this will recur on the next unrelated PR too.

makeavish (approved):

PR #3038, opened 2026-04-07
Add April 2026 404 redirect remediations

yuvraajsj18:
https://www.notion.so/signoz/Categorize-and-fix-404-URLs-from-GSC-335fcc6bcd1981128ea3c5b4bb573061?source=copy_link

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -1518,11 +1518,6 @@ module.exports = () => {
           permanent: true,
         },
         {
-          source: '/docs/faq',
-          destination: '/docs/faqs/',
-          permanent: true,
-        },
-        {
           source: '/docs/dashboards/dashboard-templates/redis',
           destination: '/docs/dashboards/dashboard-templates/overview/',
           permanent: true,
@@ -1569,7 +1564,7 @@ module.exports = () => {
         },
         {
           source: '/docs/userguide/trac',
-          destination: '/docs/instrumentation/overview/',
+          destination: '/docs/userguide/traces/',
           permanent: true,
         },
         {
@@ -1585,17 +1580,6 @@ module.exports = () => {
         {
           source: '/heroku-alternatives',
           destination: '/comparisons/heroku-alternatives/',
-          permanent: true,
-        },
-        {
-          source:
-            '/docs/openrouter-observability/\\)\u0e41\u0e25\u0e30\u0e2a\u0e32\u0e21\u0e32\u0e23\u0e16\u0e2a\u0e23\u0e49\u0e32\u0e07\u0e41\u0e14\u0e0a\u0e1a\u0e2d\u0e23\u0e4c\u0e14\u0e19\u0e35\u0e49\u0e02\u0e36\u0e49\u0e19\u0e21\u0e32\u0e44\u0e14\u0e49',
-          destination: '/docs/openrouter-observability/',
-          permanent: true,
-        },
-        {
-          source: '/docs/openrouter-observability/\\)y',
-          destination: '/docs/openrouter-observability/',
           permanent: true,
         },
         {
@@ -1666,12 +1650,6 @@ module.exports = () => {
           permanent: true,
         },
         {
-          source:
-            '/docs/anthropic-monitoring/\\)\u0e41\u0e25\u0e30\u0e2a\u0e32\u0e21\u0e32\u0e23\u0e16\u0e17\u0e33\u0e41\u0e14\u0e0a\u0e1a\u0e2d\u0e23\u0e4c\u0e14\u0e44\u0e14\u0e49',
-          destination: '/docs/anthropic-monitoring/',
-          permanent: true,
-        },
-        {
           source: '/docs/logs-management/send-logs/docker-logs/',
           destination: '/docs/userguide/collect_docker_logs/',
           permanent: true,
@@ -1687,11 +1665,6 @@ module.exports = () => {
           permanent: true,
         },
         {
-          source: '/docs/openrouter-observability/\\)gefolgt',
-          destination: '/docs/openrouter-observability/',
-          permanent: true,
-        },
-        {
           source: '/docs/configuration/deep_storage',
           destination: '/docs/logs-management/long-term-storage/',
           permanent: true,
@@ -1702,7 +1675,7 @@ module.exports = () => {
           permanent: true,
         },
         {
-          source: '/justifying-a-million-dollar-observability-bill',
+          source: '/justifying-a-million-dollar-observability-bill/',
           destination: '/blog/justifying-a-million-dollar-observability-bill/',
           permanent: true,
         },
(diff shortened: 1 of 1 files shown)


next.config.js
```diff
@@ -1426,6 +1426,1011 @@ module.exports = () => {
...
+        },
+        {
+          source:
+            '/docs/anthropic-monitoring/\\)\u0e41\u0e25\u0e30\u0e2a\u0e32\u0e21\u0e32\u0e23\u0e16\u0e17\u0e33\u0e41\u0e14\u0e0a\u0e1a\u0e2d\u0e23\u0e4c\u0e14\u0e44\u0e14\u0e49',
+          destination: '/docs/anthropic-monitoring/',
+          permanent: true,
+        },
+        {
+          source: '/docs/logs-management/send-logs/docker-logs/',
+          destination: '/docs/metrics-management/docker-container-metrics/',
```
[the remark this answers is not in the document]
yuvraajsj18: Okay fixing

next.config.js
```diff
@@ -1426,6 +1426,1013 @@ module.exports = () => {
...
+          destination: '/docs/introduction/',
+          permanent: true,
+        },
+        {
+          source: '/docs/dashboards/dashboard-templates/aws-rds/',
+          destination: '/docs/dashboards/dashboard-templates/overview/',
+          permanent: true,
+        },
+        {
+          source: '/docs/faq',
```
manika-signoz: this url does not have trailing slash, confirming if intentional

next.config.js
```diff
@@ -1426,6 +1426,1013 @@ module.exports = () => {
...
+          destination: '/docs/faqs/',
+          permanent: true,
+        },
+        {
+          source: '/docs/metrics-management/hostmetrics/',
+          destination: '/docs/infrastructure-monitoring/overview/',
+          permanent: true,
+        },
+        {
+          source: '/docs/userguide/trac',
```
manika-signoz, line 1566: this looks like a hallucinated / typo url, how frequently do we get this? not sure if this should be added

next.config.js
```diff
@@ -1426,6 +1426,1013 @@ module.exports = () => {
...
+          permanent: true,
+        },
+        {
+          source: '/heroku-alternatives',
+          destination: '/comparisons/heroku-alternatives/',
+          permanent: true,
+        },
+        {
+          source:
+            '/docs/openrouter-observability/\\)\u0e41\u0e25\u0e30\u0e2a\u0e32\u0e21\u0e32\u0e23\u0e16\u0e2a\u0e23\u0e49\u0e32\u0e07\u0e41\u0e14\u0e0a\u0e1a\u0e2d\u0e23\u0e4c\u0e14\u0e19\u0e35\u0e49\u0e02\u0e36\u0e49\u0e19\u0e21\u0e32\u0e44\u0e14\u0e49',
```
manika-signoz: what does this source url match?

next.config.js
```diff
@@ -1426,6 +1426,1013 @@ module.exports = () => {
...
+          destination: '/docs/logs-management/long-term-storage/',
+          permanent: true,
+        },
+        {
+          source: '/guides/claude\u2011api\u2011latency/',
+          destination: '/guides/claude-api-latency/',
+          permanent: true,
+        },
+        {
+          source: '/justifying-a-million-dollar-observability-bill',
```
manika-signoz: the source url does not have a trailing slash, checking if intentional

next.config.js
```diff
@@ -1426,6 +1426,1013 @@ module.exports = () => {
...
+          destination: '/docs/frontend-monitoring/',
+          permanent: true,
+        },
+        {
+          source: '/work',
+          destination: 'https://jobs.ashbyhq.com/SigNoz',
+          permanent: true,
+        },
+        {
+          source: '/docs/operate/migration/upgrade-0\\.112',
```
manika-signoz, line 1693: we do not have an upgrade guide for 0.112 - this should redirect to standard upgrade guide imo if we do not have an upgrade guide https://signoz.io/docs/operate/migration/upgrade-standard/ 

@makeavish for input

next.config.js
```diff
@@ -1426,6 +1426,1013 @@ module.exports = () => {
...
+        {
+          source: '/openfeature',
+          destination: '/blog/openfeature/',
+          permanent: true,
+        },
+        {
+          source: '/can-signoz-handle-large-scale-production-environments-effectively',
+          destination: '/faqs/can-signoz-handle-large-scale-production-environments-effectively/',
+          permanent: true,
+        },
```
manika-signoz: https://github.com/SigNoz/signoz.io/blob/84b3a625b3a2d089364512014eaa3eee62f60305/next.config.js#L2046-L2050

this is similar to the above section, trailing slash difference in source url, we should retain one or keep both variations together for maintainability

next.config.js
```diff
@@ -1426,6 +1426,1013 @@ module.exports = () => {
...
+        {
+          source: '/blog/sig-noz-hiring-across-roles',
```
manika-signoz: this has a typo in signoz, we should not add this to redirects imo, but this is non blocking

next.config.js
```diff
@@ -1426,6 +1426,1013 @@ module.exports = () => {
...
+          source: '/open-source-apm-tools',
+          destination: '/blog/open-source-apm-tools/',
+          permanent: true,
+        },
+        {
+          source: '/how-signozs-integration-with-opentelemetry-simplifies-instrumentation/',
+          destination:
+            '/faqs/how-signozs-integration-with-opentelemetry-simplifies-instrumentation/',
+          permanent: true,
+        },
```
manika-signoz, line 2290: similar to https://github.com/SigNoz/signoz.io/pull/3038/changes#r3051726328

next.config.js
```diff
@@ -1426,6 +1426,1013 @@ module.exports = () => {
...
+        {
+          source: '/docs/dashboards/dashboard-templates/claude-agent-dashboard/',
+          destination: '/docs/dashboards/dashboard-templates/claude-agent-sdk-dashboard/',
+          permanent: true,
+        },
+        {
+          source: '/docs/logs-management/send-logs/deno-logs/',
+          destination: '/docs/logs-management/send-logs-to-signoz/',
```
manika-signoz: suggestion: https://signoz.io/docs/instrumentation/opentelemetry-deno/ maybe this in destination since related to deno ? @makeavish

next.config.js
yuvraajsj18: Updated this to keep only the trailing-slash source variant.

next.config.js
yuvraajsj18: Retargeted this to `/docs/userguide/traces/`. This exact source showed up once in the GSC export, so I kept it but tightened the destination.

next.config.js
yuvraajsj18: Removed these malformed `\\)` bot-noise redirects from the batch.

next.config.js
yuvraajsj18: Updated this to keep only the trailing-slash source variant.

next.config.js
yuvraajsj18: Updated this to `/docs/operate/migration/upgrade-standard/`.

next.config.js
yuvraajsj18: Cleaned this up by removing the no-slash duplicate and keeping the trailing-slash variant only.

next.config.js
yuvraajsj18: Removed this redirect from the batch.

next.config.js
yuvraajsj18: Applied the same cleanup here and kept only the trailing-slash source variant.

next.config.js
yuvraajsj18: makes sense

manika-signoz (approved):

ankit01-oss (approved):

PR #3186, opened 2026-04-28
feat(content): remove content layer for our own solution

H4ad:
## Summary

Closes https://github.com/SigNoz/engineering-pod/issues/4666

This removes the content layer and adds an alterantive that behaves the same way.

This new plugin is under `lib`, the configuration for the collections is moved from `contentlayer.config.ts` to `lib/content-pipeline/schema.ts`.

Instead of `.contentlayer`, the new content is stored at `.content`. It supports cache (to avoid recompile again on change), and auto-refresh in case of change while developing.

I removed the usage of `rehype-plugin-minify` since it was just slowing down the build and not introducing any benefit.

Now, when running dev, it will also spawn another service to listen the mdx files changes and recompile if something changed.

When building, it will run the script of `tsx scripts/build-content.mts` to build the content first, it skips if the content didn't change to speedup the build as well. You can ignore it by passing flag `--force`.

Here's some scripts that I used to automate the verification of differences in the prod and this change.

- [compare-pages.py](https://github.com/user-attachments/files/27207615/compare-pages.py)
- [compare-pages-sync.py](https://github.com/user-attachments/files/27207616/compare-pages-sync.py)
- [visual-investigation.py](https://github.com/user-attachments/files/27207617/visual-investigation.py)
- [compare-structured-data.py](https://github.com/user-attachments/files/27207618/compare-structured-data.py)

About structured data, the only new field I added is `articleSection` with list of tags, ref: https://unhead.unjs.io/docs/schema-org/api/schema/article#types


app/rss/rssUtils.ts
```diff
@@ -59,7 +61,12 @@ const mapOpentelemetryEntries = (opentelemetries: MDXContentApiResponse | undefi
 
 export const loadPublishedPosts = async () => {
   const deploymentStatus = getDeploymentStatus()
-  const { faqs, opentelemetries, comparisons } = await fetchAllCMSContent(deploymentStatus)
+  const [{ faqs, opentelemetries, comparisons }, allBlogs, allDocs, allGuides] = await Promise.all([
+    fetchAllCMSContent(deploymentStatus),
+    getAllBlogs(),
+    getAllDocs(),
+    getAllGuides(),
+  ])
```
manika-signoz, line 69: resolvePostPath function in rssFeed.mjs determines URL prefixes via post._raw?.sourceFilePath (e.g., "blog/my-post.mdx" → /blog/my-post), this pipeline produces documents with _file instead of _raw, so sourceFilePath is always undefined. Every blog, doc, and guide post is falling through to the default /${post.slug}, generating incorrect RSS links - e.g., https://signoz.io/my-post instead of https://signoz.io/blog/my-post, please check this 

Image

app/tags/[tag]/page.tsx
```diff
@@ -11,6 +11,26 @@ import { notFound } from 'next/navigation'
...
+    return 0
+  })
+}
+
+/**
+ * Filter out drafts and return core content fields.
+ * Local implementation that doesn't require _raw field like pliny's allCoreContent.
+ */
+function filterAndMapPosts(posts: BlogMeta[]): CoreContent<Blog>[] {
+  return posts.filter((post) => !post.draft).map((post) => post as unknown as CoreContent<Blog>)
```
Comment (manika-signoz) [inline app/tags/[tag]/page.tsx]: risky typecasting, can it be typecasted to CoreContent<Blog> directly? can we check if there is a missing field here or an incompatible type which requires casting to unknown first?

H4ad:
On hold after talking to @manika-signoz , we will prioritize moving content to CMS instead of having on build time.

PR #3250, opened 2026-05-08
chore: remove slug from guides, migrate assets

manika-signoz:
1. removes slug and id fields from guides frontmatter to match cms fields
2. move assets used in guides from public folder to data-assets folder for sync


makeavish (dismissed):
Approved.

makeavish (approved):
Approved.

PR #3272, opened 2026-05-12
update: Updated the long term storage and query builder docs

LuffySama-Dev:
Updated the long term storage and query builder docs


revmag:
@LuffySama-Dev please close this if not relevant

PR #3285, opened 2026-05-13
[codex] docs: update service account role assignment

revmag:
## What changed

- Updated Service Accounts docs to say service accounts can be assigned one or more roles.
- Renamed the step from singular role assignment to role assignment in plural.
- Added least-privilege guidance for assigning only the roles required by the automation or integration.

## Source product PR

SigNoz/signoz#11231

## Validation

- yarn check:docs-metadata
- yarn test:docs-metadata
- yarn check:doc-redirects
- yarn test:doc-redirects
- git diff --check

### DIFF
--- a/data/docs/manage/administrator-guide/iam/service-accounts.mdx
+++ b/data/docs/manage/administrator-guide/iam/service-accounts.mdx
@@ -1,5 +1,5 @@
 ---
-date: 2026-04-15
+date: 2026-05-13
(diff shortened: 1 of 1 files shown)


revmag:
Closing this because the proposed docs change is inaccurate: service accounts cannot currently be assigned multiple roles in the product UI. The source PR was misinterpreted during docs triage.

PR #3291, opened 2026-05-13
perf(site): skip static build for preview/staging

H4ad:
## Summary

The initial assumption was the static build could slowdown the build by a lot, but seems like it only changes ~20s in the build time.


H4ad:
Closing since the experiment was not successfull.

PR #3303, opened 2026-05-15
chore: update deps and package

tewarig:
this fixes https://github.com/SigNoz/engineering-pod/issues/4927

  Adds targeted resolutions to package.json to drop yarn audit from 170 → 14 vulnerabilities. The 14 remaining are all in next itself,
  patched in >=15.5.16, and can only be fixed by upgrading to Next 15.

  ## Summary

  
  │ Severity │ Before │    After     │

  │ High     │ 74     │ 5 (all next) │
  │ Moderate │ 91     │ 7 (all next) │
  │ Low      │ 5      │ 2 (all next) │

  ## What's pinned

  glob, minimatch, lodash, @sentry/browser, qs, yaml, postcss, brace-expansion (bumped above the CVE patch line), plus four nested
  */js-yaml resolutions (eslint, eslintrc, cosmiconfig, pliny — all pin vulnerable 4.1.0).

  Each pin was verified by empirical removal + re-audit, not semver reasoning. Removing any of them brings back specific vulns.

 ##  Notes

  - Earlier attempt to upgrade Next 14 → 15 was reverted (commit f70a7e7ee) — Next 15 should have a different PR
 and should be tested properly 

  - The 14 remaining next advisories (image optimizer DoS, RSC request smuggling, middleware bypass, etc.) are accepted risk until the
  Next 15 upgrade lands.

  ## Test plan

  - yarn install clean, yarn audit reports 14 (all next)
  - Homepage, /api-reference/[version], login flow, and a docs/blog page render
  
  🟢  tested


app/(site)/(opentelemetry-hub-routes)/comparisons/[...slug]/page.tsx
```diff
@@ -25,14 +25,13 @@ const layouts = {
   ComparisonsLayout,
 }
 
-export const revalidate = CMS_REVALIDATE_INTERVAL
+export const revalidate = 86400 // 1 day — see CMS_REVALIDATE_INTERVAL
```
Comment (tewarig) [inline app/(site)/(opentelemetry-hub-routes)/comparisons/[...slug]/page.tsx]: A few folks might ask why these aren't using cacheLife({ revalidate: CMS_REVALIDATE_INTERVAL }) and Cache Components, since that would let us keep importing the constant. 
   
   Short answer: out of scope for this PR.                                     
  cacheLife isn't a drop-in replacement for export const revalidate — it changes the caching unit from "the route segment" to  "the data-fetching function," which means moving 'use cache' + cacheLife() + cacheTag() onto every CMS data helper        
  (fetchAllComparisonsForPage, getOpenTelemetryHubContentLayerArticles, fetchMDXContentByPath, etc.) and removing the segment 
  exports. It also requires .useCache in next.config.js, which changes prerendering behavior site-wide v
  
 its   should have its own PR with its own review surface. Bundling it here would balloon the diff and conflate a security patch with an architectural migration.                                                                                                    
                                                                                                                             
or now: inline literal + JSDoc on the constant + // see CONSTANT_NAME pointer at each call site.

we can plan a pr later for that.

tewarig:
> **Claude finished @tewarig's task in 4m 45s** —— [View job](https://github.com/SigNoz/signoz.io/actions/runs/25919199722)
> 
> ### Review: chore: update deps and package
> * [x]  Gather context and read changed files
> * [x]  Review async params/searchParams migration across all 30+ page components
> * [x]  Review middleware, route handlers, and utility changes
> * [x]  Review package.json resolutions and dependency versions
> * [x]  Post review findings
> 
> ### Summary
> The Next.js 14 → 15 migration is well-executed overall. The `@next/codemod` output is correct, `revalidate` exports are properly inlined to literals, `req.ip` removal is clean, and the `IncidentCostGraphicClient` extraction is the right pattern. Vercel preview deployed successfully.
> 
> ### Findings
> #	Severity	File(s)	Issue
> 1	low	12 files (see [inline comment](https://github.com/SigNoz/signoz.io/pull/3303#discussion_r3248373409))	**Unused `CMS_REVALIDATE_INTERVAL` imports.** After inlining `revalidate = 86400`, the constant import is dead code in 12 files. Lint didn't catch it (likely no `no-unused-imports` rule).
> 2	low	`opentelemetry/page.tsx:48` ([inline](https://github.com/SigNoz/signoz.io/pull/3303#discussion_r3248373621))	**Plain object passed where `Promise<params>` expected.** `BlogArticlePageWithOptions params={LANDING_PARAMS}` works at runtime (`await` on non-thenable returns it), but relies on `as any` to hide the type mismatch. `Promise.resolve(LANDING_PARAMS)` would be consistent with line 17.
> 3	info	`rss/[...segments]/route.ts` ([inline](https://github.com/SigNoz/signoz.io/pull/3303#discussion_r3248372255))	**Split indirection for revalidation value.** `revalidate` is now a literal `86400`, but `CACHE_CONTROL_HEADER` still references the `CMS_REVALIDATE_INTERVAL` constant. Same value today, but divergence risk if the constant changes.
> 4	info	`growthbookServer.ts` ([inline](https://github.com/SigNoz/signoz.io/pull/3303#discussion_r3248372038))	**`UnsafeUnwrapped*` casts on deprecation clock.** Already noted as follow-up in the PR — migration path is making `resolveGrowthBookAnonymousId` async.
> 5	info	—	**Next 15 default caching change.** `fetch()` calls without explicit `cache`/`next.revalidate` options are no longer cached by default. Segment-level `revalidate` exports are preserved, but ad-hoc server-side `fetch()` calls may see increased origin traffic. Worth a grep if you notice latency changes after deploy.
> ### Checklist
> * [x]  Async `params`/`searchParams` migration — all page components and route handlers correctly updated
> * [x]  `revalidate`/`dynamicParams` exports — all inlined to literals (Next 15 segment-config analyzer satisfied)
> * [x]  `req.ip` removal — clean fallback to headers
> * [x]  `IncidentCostGraphicClient.tsx` — correct `'use client'` + `dynamic()` extraction
> * [x]  `middleware.ts` — no other Next 15 breaking changes missed
> * [x]  Resolutions — `pliny/js-yaml` scoped correctly to avoid breaking `gray-matter`
> * [x]  Manual dev-server smoke test — noted as pending in PR description
> 
> No blocking issues. The unused imports (#1) are worth cleaning up but non-urgent.

have tested in local seems to run fine.

utils/growthbookServer.ts
```diff
@@ -16,7 +21,7 @@ function resolveGrowthBookAnonymousId(forceAnonymousId?: string) {
     return forceAnonymousId
   }
 
-  const requestHeaders = headers()
+  const requestHeaders = headers() as unknown as UnsafeUnwrappedHeaders
```
manika-signoz: risky typecasting, please validate behaviour or add comment if temporary

tewarig: To unblock the upgrade refactor, Next 15 ships an escape hatch type: UnsafeUnwrappedHeaders (and UnsafeUnwrappedCookies).
   Casting through it tells TypeScript "treat this Promise as if it were the unwrapped sync value." At runtime, Next 15 still allows the
   synchronous access — it just logs a deprecation warning — so the cast keeps the code working as a temporary bridge.

tewarig: anyway target is next 16 so i don't want to follow api of next 15.

manika-signoz (approved):

PR #3317, opened 2026-05-17
chore: next upgrade

tewarig:
Upgrades Next 14.2.35 → 15.5.18 and React 18.3.1 → 19.2.6, closing 14 next security advisories. Also bumps                 
  @stoplight/elements 8 → 9, recharts 2 → 3, react-medium-image-zoom 5.2 → 5.4, @types/react/@types/react-dom to 19,
  eslint-config-next and @next/* to 15. Drops unused next-image-zoom.                                                      
                                                                                                                             
  Code: async params/cookies/headers codemod (23 files), React-19 types codemod (7 files), inlined Next 15 segment-config  
  literals, removed swcMinify + req.ip, <OpenAPISpec> switched to next/dynamic({ ssr: false }), <SpeedInsights /> gated to   
  prod, RegionContext env guard. Resolutions trimmed 14 → 9. Audit: 0 vulnerabilities. Already tested.
                                                                                                                             
  Follow-ups    / need to check                                                                                                            
                                                                                                                             
  - Replace @stoplight/elements  still emits React-19 strict-mode dev warnings ( key in
   spread, useInsertionEffect) and a ReactDOM.render import warning. Production silent, restores SSR for /api-reference/* (currently client-only).                                                                                                                 
  - MDXComponents cast — unknown as MDXComponents workaround for @types/mdx@2.0.13 predating React 19's async-FC return type.
   Remove when upstream ships.                                                                                               
  - Pre-existing MDX Image not defined warnings during static gen — not caused by this PR; falls back to raw MDX render.
  Worth tracking separately.            
  
  
  testing -  
  
  for markdown use -
  following url - 
  https://signoz-web-git-chore-nextupgrade-signoz.vercel.app/blog/high-cardinality-data/ 
  https://signoz-web-git-chore-nextupgrade-signoz.vercel.app/blog/clickstack-managed-pricing-compute-costs/
  
  i have done most of the testing. 
  
  will spin agent browser as suggested by @H4ad 
  
   also did some basing testing with caching will do some more @manika-signoz  and update 
   
   1. So added new blog and removed them using the invalited end point seems to work fine.


H4ad (approved):
Let's wait approval from @manika-signoz as well, and then we will be good to merge

app/(site)/agent-native-observability/page.tsx
```diff
@@ -1,7 +1,7 @@
 import AgentNativeObservabilityPage from './AgentNativeObservabilityPage'
 import { Metadata } from 'next'
 
-export const revalidate = 60 * 60 * 24 * 365
+export const revalidate = 31536000
```
manika-signoz, line 5: nit: please add comment saying this is 1 year

app/(site)/(opentelemetry-hub-routes)/comparisons/[...slug]/page.tsx
```diff
@@ -25,14 +24,13 @@ const layouts = {
   ComparisonsLayout,
 }
 
-export const revalidate = CMS_REVALIDATE_INTERVAL
+export const revalidate = 86400
```
Comment (manika-signoz) [inline app/(site)/(opentelemetry-hub-routes)/comparisons/[...slug]/page.tsx:28]: nit: // 1 day — see CMS_REVALIDATE_INTERVAL

app/(site)/(opentelemetry-hub-routes)/guides/[...slug]/page.tsx
```diff
@@ -28,14 +27,13 @@ const layouts = {
   GuidesLayout,
 }
 
-export const revalidate = CMS_REVALIDATE_INTERVAL
+export const revalidate = 86400
```
Comment (manika-signoz) [inline app/(site)/(opentelemetry-hub-routes)/guides/[...slug]/page.tsx:31]: nit: // 1 day — see CMS_REVALIDATE_INTERVAL

app/(site)/clickstack-alternative/page.tsx
```diff
@@ -1,7 +1,7 @@
 import ClickStackAlternativePage from './ClickStackAlternativePage'
 import { Metadata } from 'next'
 
-export const revalidate = 60 * 60 * 24 * 365
+export const revalidate = 31536000
```
manika-signoz, line 5: nit: add comment saying this is 1 year please

app/(site)/rss/[...segments]/route.ts
```diff
@@ -6,7 +6,7 @@ import { loadPublishedPosts } from '../rssUtils'
 import { CMS_REVALIDATE_INTERVAL } from '@/constants/cache'
 
 export const runtime = 'nodejs'
-export const revalidate = CMS_REVALIDATE_INTERVAL
+export const revalidate = 86400
```
Comment (manika-signoz) [inline app/(site)/rss/[...segments]/route.ts:10]: nit: // 1 day — see CMS_REVALIDATE_INTERVAL

app/(site)/rss/route.ts
```diff
@@ -5,7 +5,7 @@ import { loadPublishedPosts } from './rssUtils'
 import { CMS_REVALIDATE_INTERVAL } from '@/constants/cache'
 
 export const runtime = 'nodejs'
-export const revalidate = CMS_REVALIDATE_INTERVAL
+export const revalidate = 86400
```
manika-signoz, line 9: nit: // 1 day — see CMS_REVALIDATE_INTERVAL

app/layout.tsx
```diff
@@ -92,7 +95,7 @@ export default function RootLayout({ children }: { children: React.ReactNode })
       <link rel="alternate" type="application/rss+xml" href="/feed.xml" />
 
       <body className="pl-[calc(100vw-100%)] text-white antialiased">
-        <SpeedInsights />
+        {process.env.NODE_ENV === 'production' && <SpeedInsights />}
```
manika-signoz: use VERCEL_ENV please, it is exposed by vercel 
also please check if this is intentional, why have we removed this to only render when env is production?

middleware.ts
```diff
@@ -68,8 +68,7 @@ export function middleware(req: NextRequest) {
   }
 
   const referer = req.headers.get('referer') || req.headers.get('referrer') || 'direct'
-  const ip =
-    req.ip || req.headers.get('x-forwarded-for') || req.headers.get('x-real-ip') || 'unknown'
+  const ip = req.headers.get('x-forwarded-for') || req.headers.get('x-real-ip') || 'unknown'
```
manika-signoz, line 71: req.ip was removed, is this intentional ?

components/Changelog/Title/ChangelogTitle.tsx
```diff
@@ -10,7 +10,7 @@ interface ChangelogTitleProps {
 
 const ChangelogTitle: React.FC<ChangelogTitleProps> = ({ title, link }) => {
   const [isCopied, setIsCopied] = useState<boolean>(false)
-  const timeout = useRef<ReturnType<typeof setTimeout>>()
+  const timeout = useRef<ReturnType<typeof setTimeout>>(undefined)
```
manika-signoz: can we change this to null please? undefined might be confusing, using null would clearly indicate that the state was set by dev to null initially

components/OpenAPISpec/index.tsx
```diff
@@ -1,30 +1,17 @@
 'use client'
```
manika-signoz, line 1: can we please remove the 'use client' directive here, if possible, since Inner has the directive, helps clarify boundaries

app/(site)/(opentelemetry-hub-routes)/opentelemetry/page.tsx
manika-signoz: this is an unused import now, i think, can we please remove this if unused?

app/(site)/(opentelemetry-hub-routes)/comparisons/[...slug]/page.tsx
Comment (manika-signoz) [reply app/(site)/(opentelemetry-hub-routes)/comparisons/[...slug]/page.tsx:28]: also, the constant file hasn't been changed in the PR i think but it has the interval constants exported which might confuse contributors where someone might think that changing the interval in the constants file would change the interval in all places, but it has been inlined, a comment on the files exporting the constant would be helpful, something like:
```
  // NOTE: segment-config `export const revalidate` values are inlined as literals
  // (required by Next 15 static analysis). This constant is only used at runtime.
  export const CMS_REVALIDATE_INTERVAL = 86400                                                           
```

manika-signoz:
don't see any code change related to pricing calculator, but preview and production have order disparity in the pie chart
after
Screenshot 2026-05-19 at 00 37 55
before
Screenshot 2026-05-19 at 00 38 14

middleware.ts
tewarig: yes. next 15 have removed req.ip ,  https://nextjs.org/docs/pages/api-reference/functions/next-request

app/layout.tsx
tewarig: this runs only for prod. env. in dev it will just throw error. 🤔  we can just keep it as it was before.

tewarig: we should do this cleanup in some other pr

components/OpenAPISpec/index.tsx
tewarig: no we can't do that  
we can't use ssr with dynamic in server components .

manika-signoz: got it

manika-signoz (approved):

therealpandey (approved):

PR #3747, opened 2026-07-20
docs: document logs missing-key fallback behavior

revmag:
## Summary

Documents the logs query builder change from [SigNoz/signoz#12134](https://github.com/SigNoz/signoz/pull/12134): queries referencing a field key absent from metadata no longer fail with a 400 error. They run against raw log data and return a `key not found in metadata` warning alongside results. Related behavior change: absent attribute map values (string, number, boolean) now return `NULL` in group-by and select instead of empty string.

### Changes
- **`querying-logs.mdx`** — new `#### Missing attribute keys` subsection under Filtering (fallback behavior, exact warning string, bare vs. qualified key resolution, note that the warning is API-response-only with no UI element). Added a NULL-behavior warning under Group By covering both the `null` bucket/column change and the breaking `!= ''` filter impact (recommends `EXISTS` / `NOT EXISTS` instead). Covers deliverable features 1, 2, 4, 5.
- **`search-troubleshooting.mdx`** — updated the `key not found` carve-out to state the error no longer applies to **logs or traces** (previously traces-only in the pending PR below), and now applies only to metrics and other signals.
- Updated `date` frontmatter on both edited files.

### Coordination / notes
- **Overlaps draft PR #3728** (traces missing-key fallback, source PR #12091). Both PRs edit `search-troubleshooting.mdx` at the same location. This PR's carve-out reflects the combined end state (logs **and** traces); whichever merges second will need a trivial conflict resolution there.
- **No screenshots** — the warning is returned only in the query API response body; there is no distinct UI element (banner/tooltip/header), so screenshot capture is N/A per the deliverable.
- **Not documented (out of scope):** the `EXISTS` predicate now emitting a literal `mapContains(...)` with no bind argument (deliverable feature 6) is invisible in the UI and only affects generated-SQL bind-argument inspection. No raw-SQL-export/bind-args docs page exists, so no user-facing change was warranted. Flagging for maintainer.

Source PRs: #12134

Auto-generated by docs-sync pipeline


data/docs/logs-management/querying-logs.mdx
```diff
@@ -41,6 +58,15 @@ Log queries support a full set of [aggregation functions](https://signoz.io/docs
...
 
+When you group by or select an attribute key that is not present on every log, logs missing that key now appear in a `null` group or column instead of an empty-string (`''`) group. This makes attribute maps (string, number, boolean) consistent with the existing behavior for resource attributes.
+
+<Admonition type="warning">
+This is a breaking change for filters and rules that rely on the old empty-string behavior:
+
+- **Group-by matching:** If you have dashboards or alert rules that match or compare against the empty-string group-by bucket, update them to match `null` instead. Matching on `''` now silently returns no rows for logs that lack the grouped key.
+- **`!= ''` filters:** A filter such as `mykey != ''` used to exclude logs where the attribute was absent. Absent values now return `null`, and `null != ''` evaluates to `null` (falsy), so those logs are no longer matched. To exclude logs that lack a key, use `mykey EXISTS`; to include only logs missing a key, use `NOT mykey EXISTS`.
+</Admonition>
+
```
tushar-signoz, line 69: Remove this part please, it not correct

tushar-signoz (changes requested):

PR #3814, opened 2026-07-24
docs: add dashboards v2 schema as a public resource for old MCP servers

namanv3:
## Pull Request

---

### 📄 Summary
 
Users could be using an old version of the MCP server while their signoz server moves on to the Perses release. In such a scenario those MCP servers will hit the V1 API and get an error. That error message should contain a public facing link for the dashboards schema. The files added in this PR act as those public facing resources


namanv3:
discussed that we don't need this

PR #3872, opened 2026-07-31
feat(dashboards): offer V2 dashboard template JSON and deprecate V1

AshwinBhatkal:
## Pull Request

---

### 📄 Summary

[SigNoz/dashboards#396](https://github.com/SigNoz/dashboards/pull/396) introduces a V2 (Perses) JSON for every dashboard: the V2 file takes the existing root path, and the legacy V1 copy moves into a `v1/` subdirectory. Template pages currently expose exactly one download and say nothing about versions, so after that PR merges every page would silently hand out V2 JSON — which fails to import on SigNoz older than v0.135.0 — with no way to reach V1 and no warning either way.

This makes the version explicit and steers readers to V2.

`DashboardActions` becomes a **"Dashboard JSON" card** with a version segmented control. Version is a *mode*, not an action, so it is a tab pair rather than four buttons — there are still only copy and download. The deprecation warning appears only while V1 is selected, i.e. exactly when the reader is considering it, instead of sitting on every page as permanent noise. The V1 download also drops from primary to secondary so the visual weight matches the recommendation.

The V1 tab stays visible rather than hidden behind a disclosure: users on older self-hosted SigNoz genuinely need it and should not have to hunt. It is labelled `V1 (legacy)` and de-emphasized instead. Templates with no V1 file render with no version chrome at all.

Wording is precise about both directions of incompatibility: V2 **requires** v0.135.0+, while V1 still imports on newer versions through the temporary server-side conversion shim that will be removed.

#### Screenshots / Screen Recordings (if applicable)

https://github.com/user-attachments/assets/d7cdca65-4f66-4d28-94f4-4c854c2844e3

#### Issues closed by this PR

N/A

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🐛 Bug Context

Three pre-existing `DashboardActions` bugs are fixed alongside the main change, since it rewrites the same code paths.

#### Root Cause

- `dashboardName=""` was passed by 36 pages. The download name was built as `` `${dashboardName.toLowerCase()...}.json` ``, so an empty name produced a bare `.json` — a hidden dotfile.
- Fetch failures called `alert()`, which is jarring and offers no recovery path.
- Copy and download each refetched the JSON independently, so copy-after-download hit the network twice.

#### Fix Strategy

File names fall back to the JSON basename and carry a version suffix (`nginx-v2.json`, `agno-dashboard-v2.json`). Failures render an inline error with a link to the raw JSON. Fetched JSON is cached per URL for the component's lifetime.

---

### 🧪 Testing Strategy

- **Tests added/updated:** three cases in `tests/agent-markdown-stubs.test.js` covering the new `DashboardActions` agent stub — both versions present, V2-only, and the legacy `dashboardJsonUrl` alias.
- **Manual verification:** drove the real pages in Chromium against `yarn dev`, with the dashboards-repo requests stubbed so the assertions do not depend on the unmerged PR:
  - V2 download → `nginx-v2.json` with V2 content; V2 copy → V2 content on the clipboard
  - switch to V1 → deprecation notice renders, download → `nginx-v1.json` with V1 content, copy → V1 content
  - no-name page (`agno-dashboard`) → `agno-dashboard-v2.json`, confirming the dotfile fix
  - forced 404 → inline error with the raw-JSON fallback link, no `alert()`
  - V2-only page (`openai-dashboard`) → zero version tabs
  - narrow viewport (380px) → header and buttons wrap without horizontal overflow
- **Content verification against the dashboards PR branch:** all 84 generated V1 URLs return 200 on `nv/perses-dashboards`, and all 101 V2 URLs carry `schemaVersion: "v6"`.
- **Repo checks:** `check:docs-metadata`, `test:docs-metadata`, `check:doc-redirects`, `test:doc-redirects`, `check:stale-urls`, `test:stale-urls`, `yarn lint` (0 errors; 180 pre-existing warnings, none in touched files), `yarn build`, `yarn test` (56 passing).
- **Edge cases covered:** V2-only templates, V1-only (component falls back to V1 as the active version), the deprecated `dashboardJsonUrl` alias, empty `dashboardName`, fetch failure, clipboard failure.

---

### ⚠️ Risk & Impact Assessment

- **Blast radius:** 105 dashboard template docs plus the templates overview and import guide. One component, used only by these pages. No routes, slugs, or redirects change.
- **Potential regressions:** low. The component keeps accepting `dashboardJsonUrl` as a V2 alias, so any page that still passes it renders normally. The main risk is ordering, not correctness — see the blocking note below.
- **Rollback plan:** revert the PR. Nothing is stateful and no URLs move.

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | Cloud / OSS / Enterprise |
| Change Type | Feature |
| Description | Dashboard template pages now offer both V2 and V1 dashboard JSON, defaulting to V2 and marking V1 as deprecated for SigNoz v0.135.0 and newer. |

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented

---

## 👀 Notes for Reviewers

**🚧 Do not merge before [SigNoz/dashboards#396](https://github.com/SigNoz/dashboards/pull/396).** That PR is what makes the root paths serve V2 and creates the `v1/` copies. Until it lands, every URL labelled "V2" here still serves V1 content and all 84 V1 URLs 404. This is purely a sequencing constraint — I verified against the `nv/perses-dashboards` branch that all the URLs resolve and carry the expected schema.

**⚠️ dashboards#396 drops V1 for 22 dashboards.** These are modified in place with no `v1/` copy preserved:

`autogen`, `azure-openai`, `bedrock`, `crewai`, `dify`, `google-adk`, `google-gemini`, `grok`, `groq`, `hermes`, `inkeep`, `litellm` (proxy + sdk), `livekit`, `mastra`, `ollama`, `openai`, `opencode`, `pipecat`, `pydantic-ai`, `semantic-kernel`, `temporal-agents`

I have shipped those pages as V2-only, which is right if it is intentional. If it is an oversight, those users lose their V1 fallback and it is worth raising on that PR — the URLs get picked up automatically once the files exist.

**Two related surfaces left alone, flagged for a call:**

- `data/docs/integrations/temporal-cloud-metrics.mdx:258` links a raw dashboard JSON directly in prose, so it will silently become V2 with no version note. Out of scope here, but easy to fold in if you want it.
- Several `constants/listicles/dashboard-templates.json` entries link to GitHub tree pages instead of doc pages (`aws-rds`, `ecs-infra-metrics`, `jenkins`, `jmx`, and others). Readers there will now see both a `v1/` directory and the V2 file with no guidance.

**Review pointers:**

- The 105 MDX files were migrated by script and are mechanically identical: prop rename, V1 URL added where a `v1/` copy exists, wrapper `div` dropped. `data/docs/dashboards/dashboard-templates/nginx.mdx` is representative — the rest follow the same shape.
- The `date:` frontmatter bump across all 107 docs is required by `check:docs-metadata`, which rejects a modified doc whose date is more than 7 days old.
- Commits are split by concern: component → agent markdown → docs.

### DIFF
--- a/components/Dashboards/DashboardActions.tsx
+++ b/components/Dashboards/DashboardActions.tsx
@@ -1,7 +1,8 @@
 'use client'
 
 import React, { useCallback, useMemo, useRef, useState } from 'react'
-import { Check, Copy, Download, TriangleAlert } from 'lucide-react'
+import { Check, Copy, Download } from 'lucide-react'
+import Admonition from '../Admonition/Admonition'
 import Button from '../ui/Button'
 
 export const V2_MIN_SIGNOZ_VERSION = 'v0.135.0'
@@ -64,11 +65,20 @@ const TAB_CLASS = [
 
 // `not-prose` drops the docs anchor styling, so links carry it themselves.
 const LINK_CLASS = 'text-[var(--accent-primary)] underline underline-offset-2'
+// The raw JSON lives on GitHub, so following it leaves the docs site.
+const EXTERNAL_LINK_PROPS = {
+  target: '_blank',
+  rel: 'noopener noreferrer nofollow',
+} as const
 
-const V1_NOTICE_CLASS = [
-  'mb-4 flex gap-2 rounded-[4px] p-3 text-sm leading-6 text-[var(--l2-foreground)]',
-  'border border-[color-mix(in_srgb,var(--accent-amber)_35%,transparent)]',
-  'bg-[color-mix(in_srgb,var(--accent-amber)_8%,transparent)]',
+/**
+ * Low-emphasis action styling for the deprecated version. The `secondary` Button
+ * variant hardcodes dark ink/vanilla colors, so it stayed dark on a light ground;
+ * these semantic tokens flip with the theme like the rest of the card.
+ */
+const V1_BUTTON_CLASS = [
+  'border border-[var(--l1-border)] bg-transparent text-[var(--l2-foreground)]',
+  'hover:bg-[var(--l3-background)] hover:text-[var(--l1-foreground)]',
 ].join(' ')
 
 const DashboardActions: React.FC<DashboardActionsProps> = ({
@@ -192,22 +202,14 @@ const DashboardActions: React.FC<DashboardActionsProps> = ({
 
       <div className="px-4 py-4">
         {isV1 ? (
-          <p className={V1_NOTICE_CLASS}>
-            <TriangleAlert
-              aria-hidden
-              className="mt-1 h-4 w-4 shrink-0 text-[var(--accent-amber)]"
-            />
-            <span>
-              <strong className="font-semibold text-[var(--l1-foreground)]">Deprecated.</strong> Use
-              V1 only on SigNoz older than {V2_MIN_SIGNOZ_VERSION}. From {V2_MIN_SIGNOZ_VERSION}{' '}
-              onwards, import the V2 JSON — V1 imports still work through a temporary compatibility
-              path that will be removed. See the{' '}
-              <a className={LINK_CLASS} href={UPGRADE_GUIDE_URL}>
-                {V2_MIN_SIGNOZ_VERSION} upgrade guide
-              </a>
-              .
-            </span>
-          </p>
+          // Admonition already carries the warning tone in both themes, so the
+          // notice does not hand-roll its own amber surface.
+          <Admonition type="warning" variant="sm" title="V1 is deprecated" className="mt-0">
+            Use V1 only on SigNoz older than {V2_MIN_SIGNOZ_VERSION}. From {V2_MIN_SIGNOZ_VERSION}{' '}
+            onwards, import the V2 JSON — V1 imports still work through a temporary compatibility
+            path that will be removed. See the{' '}
+            <a href={UPGRADE_GUIDE_URL}>{V2_MIN_SIGNOZ_VERSION} upgrade guide</a>.
+          </Admonition>
         ) : (
           <p className="mb-4 text-sm leading-6 text-[var(--l2-foreground)]">
             Recommended. Uses the{' '}
@@ -221,7 +223,8 @@ const DashboardActions: React.FC<DashboardActionsProps> = ({
         <div className="flex flex-wrap gap-3">
           <Button
             // The deprecated version should not carry the primary CTA weight.
-            variant={isV1 ? 'secondary' : 'default'}
+            variant={isV1 ? 'ghost' : 'default'}
+            className={isV1 ? V1_BUTTON_CLASS : undefined}
             rounded="default"
             isButton={true}
             onClick={handleDownload}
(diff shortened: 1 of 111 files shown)


therealpandey (approved):

namanv3 (approved):
Tested the deployment after merging the dashboards repo PR, looks good to me.

components/Dashboards/DashboardActions.tsx
```diff
@@ -1,116 +1,280 @@
...
+        )}
       </div>
-      
-      <div className="text-sm text-gray-600 mt-2">
-        <p className="text-center">
-          <span className="font-bold italic">Dashboards → + New dashboard → Import JSON</span>
+
+      <div className="px-4 py-4">
+        {isV1 ? (
+          <p className={V1_NOTICE_CLASS}>
```
manika-signoz: can use admonition warning component, already handles dark and light mode

components/Dashboards/DashboardActions.tsx
```diff
@@ -1,116 +1,280 @@
...
+
+        <div className="flex flex-wrap gap-3">
+          <Button
+            // The deprecated version should not carry the primary CTA weight.
+            variant={isV1 ? 'secondary' : 'default'}
+            rounded="default"
+            isButton={true}
+            onClick={handleDownload}
+            disabled={pending !== null}
+          >
```
manika-signoz, line 214: v1 legacy button differs in styling from v2 in light and dark mode - remove dark classname from html using devtools just to see how the component looks in light mode

Image

components/Dashboards/DashboardActions.tsx
```diff
@@ -1,116 +1,280 @@
...
+            <a className="underline underline-offset-2" href={activeUrl}>
+              Open the raw JSON
+            </a>{' '}
+            instead.
+          </p>
+        )}
+
+        <p className="mt-4 text-xs leading-5 text-[var(--l3-foreground)]">
+          Import it in SigNoz with <strong>Dashboards → + New dashboard → Import JSON</strong>.{' '}
+          <a className={LINK_CLASS} href={IMPORT_DOC_URL}>
```
manika-signoz: should open in new tab, currently opens in existing tab which might break reader flow

AshwinBhatkal: all in app links don't open in new tab 🤔

AshwinBhatkal: Use CustomLink component

components/Dashboards/DashboardActions.tsx
AshwinBhatkal: done

manika-signoz (approved):

PR #3805, opened 2026-07-23
feat: adding GCP integration docs for manual flow

swapnil-signoz:
https://staging.signoz.io/docs/integrations/gcp/gcp-integration/

## Pull Request

Adding docs for GCP integration manual flow

---

### 📄 Summary
**Why does this change exist?**
We are starting GCP integration with manual flow where users own the resources and their life cycle, in that case docs are important to guide users through each step of creating these resources and complete the ingestion pipeline.

**What problem does it solve, and why is this the right approach?**
Helping users step by step to create GCP resources to ingest data in Signoz.

#### Issues closed by this PR
https://github.com/SigNoz/platform-pod/issues/2902

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature
- [ ] 🐛 Bug fix
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

PR #3923, opened 2026-08-08
refactor(docs): remove GCP logs collection, switch to contrib collector

swapnil-signoz:
Stacked on #3805 — base is `feat/gcp-integration-docs`, so this diff shows only the removal.

Per SigNoz/platform-pod#2901 (MoM 7 Aug 2026), GCP ships **metrics-only**: the
`googlecloudpubsubpush` receiver is dev-stage and compiled into no published Collector
distribution. 

Deletes `collect-logs.mdx` and strips logs from the other 8 pages, renumbering the
flow from 5 steps to 4. 

The Collector moves to `otel/opentelemetry-collector-contrib` >= `0.158.0`, and since a metrics-only Collector opens no listener, `health_check` extension replaces the push receiver's port for Cloud Run readiness.


data/docs/integrations/gcp/deploy-otel-collector.mdx
```diff
@@ -23,28 +23,18 @@ The OpenTelemetry Collector is the engine of the GCP integration: it pulls metri
...
 
-Check the <a href="https://hub.docker.com/r/signoz/signoz-otel-collector/tags" target="_blank" rel="noopener noreferrer nofollow">tags list</a> for the latest version and use that tag (for example, `signoz/signoz-otel-collector:v0.157.0`) instead of `latest`, so deployments stay reproducible.
-
-<Admonition type="warning">
-Use **v0.157.0 or later**. Earlier versions don't include the receivers this integration depends on.
-</Admonition>
+Check the <a href="https://hub.docker.com/r/otel/opentelemetry-collector-contrib/tags" target="_blank" rel="noopener noreferrer nofollow">tags list</a> for the latest version and use that tag (for example, `otel/opentelemetry-collector-contrib:0.158.0`) instead of `latest`, so deployments stay reproducible. The version must be **0.158.0 or above** — earlier versions report cumulative GCP metrics in a way that prevents rate functions on counter metrics. Note that these tags have no `v` prefix.
```
swapnil-signoz: **Em dashes in prose (4 instances across 2 files)**

`contributing/docs-authoring.md` says to avoid em dashes (`—`) and use a period, comma, colon, or parentheses instead. This line has two.

**Impact:** Style inconsistency with the rest of the docs, and the review rubric flags em dashes as an AI-writing tell.

**Fix** for this line:

```
... so deployments stay reproducible. The version must be **0.158.0 or above**: earlier versions report cumulative GCP metrics in a way that prevents rate functions on counter metrics. Note that these tags have no `v` prefix.
```

Other instances to fix in the same pass:
- `deploy-otel-collector.mdx:83` — `on a port — and a Cloud Run`
- `deploy-otel-collector.mdx:85` — `no port listener — you lose the readiness signal`
- `collect-metrics.mdx:36` — `from the starter config — **replace only its**`

data/docs/integrations/gcp/deploy-otel-collector.mdx
```diff
@@ -92,16 +80,14 @@ service:
...
-</Admonition>
+**Why the `health_check` extension is here.** This Collector only makes outbound calls, so nothing else listens on a port — and a Cloud Run *service* marks a revision ready only once the container accepts connections on `--port`. `health_check` provides that listener, so a broken config fails the deploy instead of running silently and collecting nothing. Keep the extension's `endpoint` port and the Cloud Run `--port` the same (`8090` here).
 
-<Admonition type="info">
-**Cloud Run requires a listening port.** Cloud Run health-checks a service by connecting to the port named in `--port` and considers the revision ready only once the container listens on it. The `googlecloudpubsubpush` receiver listens on `:8090`, so this config also serves as the container's HTTP port. Keep the receiver `endpoint` port and the Cloud Run `--port` the same value (`8090` here).
+To skip the health endpoint, deploy to a <a href="https://cloud.google.com/run/docs/resource-model" target="_blank" rel="noopener noreferrer nofollow">Cloud Run worker pool</a>, which needs no port listener — you lose the readiness signal in exchange.
 </Admonition>
```
swapnil-signoz, line 88: **`health_check` overstates what it detects: a broken config does not "fail the deploy"**

Two separate mechanisms are conflated here:

1. **Config parse/validation errors** — the Collector exits non-zero at startup, the container never listens, and the Cloud Run revision fails readiness. This part is correct, and it is true with or without `health_check`.
2. **A config that parses but collects nothing** — wrong `project_id`, a `metrics_list` metric the SA can't read, a `monitoring.viewer` grant missing on a monitored project. The Collector starts fine and `health_check` returns **200 healthy**. The deploy succeeds and the service "looks healthy while collecting nothing", which is precisely the failure mode the note claims to prevent.

The extension's own README warns off the only feature that would have checked pipeline health:

> The `check_collector_pipeline` feature of this extension is not working as expected. It is recommended to not use the feature.

So `health_check` here is a **liveness listener that satisfies Cloud Run's port requirement**. It is not a collection-correctness signal.

**Impact:** A reader who trusts this note will treat a green Cloud Run revision as proof that metrics are flowing, skip the `## Validate` step, and then have no idea why dashboards are empty. That is the most likely first-run failure on this page.

**Fix** — scope the claim to what it actually does:

```md
<Admonition type="info">
**Why the `health_check` extension is here.** This Collector only makes outbound calls, so nothing else listens on a port, and a Cloud Run *service* marks a revision ready only once the container accepts connections on `--port`. `health_check` provides that listener. Keep the extension's `endpoint` port and the Cloud Run `--port` the same (`8090` here).

A ready revision means the Collector started and its config parsed. It does **not** confirm metrics are flowing: a config that parses but has the wrong `project_id` or a missing `roles/monitoring.viewer` grant still reports healthy. Confirm collection in [Collect metrics](https://signoz.io/docs/integrations/gcp/collect-metrics/).

To skip the health endpoint, deploy to a <a href="https://cloud.google.com/run/docs/resource-model" target="_blank" rel="noopener noreferrer nofollow">Cloud Run worker pool</a>, which needs no port listener (you lose the readiness signal in exchange).
</Admonition>
```

Source: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/v0.158.0/extension/healthcheckextension/README.md

data/docs/integrations/gcp/create-service-account.mdx
```diff
@@ -10,13 +10,12 @@ SigNoz needs a dedicated **service account** to act as the single identity for t
...
 
 ## Prerequisites
 
-- A GCP account with permission to enable APIs, create service accounts, and grant IAM roles in your deployment project and every project you want to monitor. The **Owner** or **Editor** role on those projects is sufficient; if you use custom roles, you need `resourcemanager.projects.setIamPolicy`, `iam.serviceAccounts.create`, and `serviceusage.services.enable`.
+- A GCP account with permission to enable APIs, create service accounts, and grant IAM roles in your deployment project and every project you want to monitor. The **Owner** or **Editor** role on those projects is sufficient; if you use custom roles, you need `resourcemanager.projects.setIamPolicy`, `iam.serviceAccounts.create`, and `serviceusage.services.enable`, plus `roles/iam.serviceAccountUser` on the service account you create here so you can attach it to the Cloud Run service in <a href="https://cloud.google.com/run/docs/deploying" target="_blank" rel="noopener noreferrer nofollow">step 3</a>.
```
swapnil-signoz: **`roles/iam.serviceAccountUser` is mixed into a list of granular permissions, and the link points at the wrong page**

The fix itself is correct and a good catch. Two problems with how it landed:

**1. Type mismatch inside one list.** The sentence reads "if you use custom roles, you need `resourcemanager.projects.setIamPolicy`, `iam.serviceAccounts.create`, and `serviceusage.services.enable`, plus `roles/iam.serviceAccountUser`". The first three are granular **permissions** you add to a custom role; the last is a **predefined role**. A reader building a custom role will look for a permission named `roles/iam.serviceAccountUser` and not find one. The underlying permission is `iam.serviceAccounts.actAs`.

**2. Scope and link are off.** The grant is needed **on the service account** (the service identity), not on the project, and it is needed by the **human deploying**, not by `signoz-integration`. The `<a href=...>step 3</a>` link goes to Google's generic Cloud Run deploying page while the surrounding text says "step 3", which in this doc set means [Deploy the Collector](https://signoz.io/docs/integrations/gcp/deploy-otel-collector/). Two different "step 3"s.

**Impact:** The custom-role path is the only path this line affects, and it is the path where a reader has no fallback. Getting `actAs` wrong produces a `PERMISSION_DENIED ... iam.serviceaccounts.actAs` error at deploy time with nothing on this page to connect it back.

**Fix:**

```md
- A GCP account with permission to enable APIs, create service accounts, and grant IAM roles in your deployment project and every project you want to monitor. The **Owner** or **Editor** role on those projects is sufficient. If you use custom roles, you need `resourcemanager.projects.setIamPolicy`, `iam.serviceAccounts.create`, and `serviceusage.services.enable` on the project, plus `iam.serviceAccounts.actAs` on the service account you create here (granted via **Service Account User**, `roles/iam.serviceAccountUser`) so you can attach it to the Cloud Run service in [step 3](https://signoz.io/docs/integrations/gcp/deploy-otel-collector/).
```

Source: https://cloud.google.com/run/docs/deploying (Required roles: `roles/run.developer` on the service, `roles/iam.serviceAccountUser` on the service identity)

data/docs/integrations/gcp/create-service-account.mdx
```diff
@@ -133,20 +127,12 @@ gcloud projects add-iam-policy-binding <monitored-project-id> \
...
-- **Push-auth service account**: the Pub/Sub push subscription authenticates to the Collector's Cloud Run service using a service account that holds `roles/run.invoker` on that service. You can reuse `signoz-integration` for this or create a dedicated one.
-
 ## Permission summary
 
 | Identity | Role | Scope | Purpose |
 | --- | --- | --- | --- |
 | `signoz-integration` service account | `roles/monitoring.viewer` | Each monitored project | Read metrics via the Monitoring API |
-| Log sink writer identity (auto-created) | `roles/pubsub.publisher` | The Pub/Sub topic | Publish routed logs (set up in [Collect logs](https://signoz.io/docs/integrations/gcp/collect-logs/)) |
-| Push-auth service account | `roles/run.invoker` | The Collector's Cloud Run service | Authenticate Pub/Sub push to the Collector (set up in [Collect logs](https://signoz.io/docs/integrations/gcp/collect-logs/)) |
+| `signoz-integration` service account | `roles/secretmanager.secretAccessor` | The Collector config secret | Read the Collector config at startup (granted in [Deploy the Collector](https://signoz.io/docs/integrations/gcp/deploy-otel-collector/)) |
```
swapnil-signoz: **Permission summary now lists a role this page never grants, with no matching step**

The new row documents `roles/secretmanager.secretAccessor` and says it is "granted in [Deploy the Collector]". But this page's own steps only ever grant `roles/monitoring.viewer` (step 3). The two rows in this table are now inconsistent in kind: one is granted here, one is granted on a different page.

The deleted rows had the same shape ("set up in Collect logs"), so this is carried-over structure rather than a new mistake. It reads worse now that only two rows remain and one of them is a forward reference.

**Impact:** A reader auditing what `signoz-integration` holds after finishing *this* page will believe `secretAccessor` is already granted and skip it in step 2 of Deploy the Collector, where it is actually required for the Collector to read its config at startup.

**Fix** — mark the scope explicitly so the table reads as the full permission model rather than as this page's output:

```md
## Permission summary

After this guide, `signoz-integration` holds:

| Identity | Role | Scope | Purpose |
| --- | --- | --- | --- |
| `signoz-integration` service account | `roles/monitoring.viewer` | Each monitored project | Read metrics via the Monitoring API |

One more role is granted later, in [Deploy the Collector](https://signoz.io/docs/integrations/gcp/deploy-otel-collector/#step-2-store-the-config-in-secret-manager):

| Identity | Role | Scope | Purpose |
| --- | --- | --- | --- |
| `signoz-integration` service account | `roles/secretmanager.secretAccessor` | The Collector config secret | Read the Collector config at startup |
```

data/docs/integrations/gcp/gcp-integration.mdx
```diff
@@ -42,39 +41,36 @@ Choose how you want to connect GCP to SigNoz:
...
 
-<div className="block overflow-hidden rounded border border-gray-700 bg-gray-900 p-6 opacity-60 dark:bg-gray-800">
-    <div className="mb-2 text-xl font-bold text-white dark:text-gray-100">SigNoz Agent (Coming soon)</div>
-    <p className="text-base text-gray-400 dark:text-gray-300">SigNoz deploys and manages the Collector for you: connect your GCP account, toggle a service on, and SigNoz provisions the metric receivers and log routing in the background.</p>
+<div className="block rounded border border-solid border-[var(--l1-border)] bg-[var(--l2-background)] p-4 opacity-60">
+    <div className="flex flex-col gap-6">
+        <div className="text-base font-medium leading-none text-[var(--l1-foreground)]">SigNoz Agent (Coming soon)</div>
+        <p className="m-0 text-[13px] leading-5 tracking-[-0.065px] text-[var(--l2-foreground)]">SigNoz deploys and manages the Collector for you: connect your GCP account, toggle a service on, and SigNoz provisions the metric receivers in the background.</p>
+    </div>
 </div>
```
swapnil-signoz, line 54: **Hand-rolled div duplicating `DocCard` styling — extract a `disabled`/`comingSoon` variant instead**

This commit hand-tunes raw Tailwind (`text-[13px]`, `tracking-[-0.065px]`, `leading-none`, `gap-6`) to match `components/DocCard.tsx`. That is a copy of a component's internals into MDX: the next change to `DocCard` silently desynchronizes this card, and the magic numbers give no hint they are meant to track another file.

The same "coming soon" pattern already appears in the `entityName="setup"` tabs on `cloud-sql-postgresql.mdx` and `memorystore-redis.mdx`, so this is a third spelling of one concept.

**Impact:** Maintenance only, no reader-facing bug today. Worth fixing now because this is the pattern other integration overviews will copy.

**Fix** — add an optional prop to `DocCard` and use the component:

```tsx
// components/DocCard.tsx
export default function DocCard({ title, description, href, comingSoon = false }) {
  if (comingSoon) {
    return (
      <div className="... opacity-60" aria-disabled="true">
        {/* same markup as the normal card, minus the link */}
      </div>
    )
  }
  // existing link card
}
```

```mdx
<DocCard
    title="SigNoz Agent (Coming soon)"
    description="SigNoz deploys and manages the Collector for you: connect your GCP account, toggle a service on, and SigNoz provisions the metric receivers in the background."
    comingSoon
/>
```

Also note the raw `<div>` carries no `aria-disabled` or equivalent, so screen readers announce it as ordinary content indistinguishable from the actionable card next to it.

If you'd rather not touch `DocCard` in a docs-content PR, that's a reasonable call — worth a follow-up issue so the duplication is tracked rather than settled by default.

swapnil-signoz:
## Docs review summary

Reviewed the logs removal and the contrib-collector swap across 8 GCP pages, the sidebar, and the redirect. The removal itself is clean: no `collect-logs` or `gcp-integration-logs-pipeline` reference survives anywhere in `data/`, `constants/`, `app/`, `components/`, or `utils/`, and the 5→4 step renumbering is consistent across `manual-setup.mdx`, `deploy-otel-collector.mdx`, and both service pages.

Findings are concentrated in the two things the PR *added* rather than removed: the `health_check` rationale and the `serviceAccountUser` prerequisite.

### Key findings

**P1**
- `deploy-otel-collector.mdx:82-86` — The `health_check` note claims a config error "surfaces as a failed deployment instead of a service that looks healthy while collecting nothing." It only catches config *parse* failures. A config that parses but has a wrong `project_id` or a missing `roles/monitoring.viewer` grant returns 200 healthy and deploys green. The extension's README explicitly warns off `check_collector_pipeline`, the only feature that would have checked collection. This is the page's most likely first-run failure and the note currently points readers away from `## Validate`.
- `create-service-account.mdx:18` — `roles/iam.serviceAccountUser` (a predefined role) is listed inside a series of granular permissions, so the custom-role reader looks for a nonexistent permission by that name; the underlying one is `iam.serviceAccounts.actAs`. Scope is also unstated (on the service account, held by the deploying human, not by `signoz-integration`), and the inline "step 3" link goes to Google's generic deploying page while the doc set's own step 3 is Deploy the Collector.

**P2**
- `create-service-account.mdx:135` — Permission summary lists `roles/secretmanager.secretAccessor` as granted, but nothing on this page grants it. Carried-over structure from the deleted log-identity rows, and it now reads as this page's output. A reader may skip the grant in step 2 of Deploy the Collector, where it is actually required.
- `gcp-integration.mdx:49-54` — Hand-rolled div reproducing `DocCard.tsx` internals (`text-[13px]`, `tracking-[-0.065px]`) with no `aria-disabled`. Suggest a `comingSoon` variant on `DocCard`; a follow-up issue is a fine alternative.

**P3**
- Em dashes at `deploy-otel-collector.mdx:30,83,85` and `collect-metrics.mdx:36`, against the authoring rule.

### Technical verification

Every load-bearing claim in the PR description checks out, and I verified the ones the docs assert rather than taking the description at face value:

- **`googlecloudpubsubpush` absent from contrib** — confirmed. `cmd/otelcontribcol/builder-config.yaml` at v0.158.0 lists `googlecloudmonitoringreceiver` and `googlecloudlogentryencodingextension` but no `googlecloudpubsubpush`. The premise for dropping logs holds.
- **0.158.0 monotonic floor** — confirmed, and it is the right floor. Contrib `CHANGELOG.md` v0.158.0: `receiver/googlecloudmonitoring: CUMULATIVE metrics from Cloud Monitoring will now be properly marked as monotonic upon conversion (#49804)`. Issue #49804 closed 2026-07-22, before the release.
- **Bare tags, no `v` prefix** — confirmed. Docker Hub returns `0.158.0` (published 2026-08-04); `v0.158.0` 404s. The doc's warning is accurate and non-obvious.
- **`health_check` in contrib, config surface valid** — confirmed. `endpoint` is a real setting and `0.0.0.0:8090` is valid; the doc correctly avoids the broken `check_collector_pipeline`. Only the *claim about what it detects* is wrong (P1 above).
- **`roles/iam.serviceAccountUser` for Cloud Run** — confirmed as mandatory. Google lists it "on the service identity" alongside `roles/run.developer`. Real pre-existing gap, correctly identified.
- `googlecloudmonitoring` config keys (`project_id`, `collection_interval`, `metrics_list`/`metric_name`) match the receiver README; `300s` default and `60s` minimum are accurate.
- OTLP export uses HTTP `:443` via `otlphttp`, per the OTLP/HTTP-over-gRPC preference. `<region>` token used correctly, so the region selector works.

Sources: `open-telemetry/opentelemetry-collector-contrib` v0.158.0 `builder-config.yaml`, `CHANGELOG.md`, `extension/healthcheckextension/README.md`, `receiver/googlecloudmonitoringreceiver/README.md`; `https://cloud.google.com/run/docs/deploying`.

### Personas and fit

Platform / cloud engineer with GCP IAM rights, setting up agentless GCP metrics. Fit is good and improved by this PR: dropping logs removes the Pub/Sub topic, push subscription, per-project sink, writer-identity grant, and `_Default` exclusion filter from the mandatory path, which was the bulk of the setup burden for a metrics-only outcome. Weakest point is the P1 `health_check` note, which sets a false expectation for exactly this persona's verification habit.

### JTBD coverage

Passing: single obvious job per page; happy path end to end; minimal mandatory steps; concrete `## Validate` on all four service/setup pages; symptom-first troubleshooting; `<region>` handling; sidebar updated; redirect added; all links absolute `signoz.io` internal / MDX-anchor external.

At risk:
- **Check 12 (validation tells users what success looks like)** — undermined on `deploy-otel-collector.mdx` by the `health_check` claim, not by a missing section.
- **Check 10 (prerequisites have a direct step)** — the `secretAccessor` row and the mislabeled `actAs` requirement both weaken the permission model a reader assembles from `create-service-account.mdx`.
- **Check 19 (clean prose)** — em dashes.

Frontmatter is in range on all 8 pages (titles 44–56 chars, descriptions 127–159). `cloud-sql-postgresql.mdx` title is 44 chars, just under the 50–60 band; low priority, and padding it for length alone is not worth it.

### Checklist coverage

`check:docs-metadata`, `check:doc-redirects` both pass locally. Images are WebP and above the 1200 px floor (architecture 2800×1200, connect-account 3252×2134, control-plane 3250×1938). No new `components/**/*.mdx` partials, so the prefer-`.tsx` guidance does not apply. Redirect follows the existing dual with/without-trailing-slash convention under `trailingSlash: true`.

### Open questions and assumptions

- **Three images still show logs.** The PR notes the architecture diagram and two UI screenshots are being updated separately. The architecture `.webp` is modified in this PR and the two screenshots show as modified in the working tree, so this may already be partly done — but I reviewed alt text and captions, not pixels, and cannot confirm the rendered diagrams no longer show Pub/Sub, the log sink, or a **Log Collection** toggle. Please confirm before merge: `connect-gcp-account.mdx` now states each service "has a **Metric Collection** toggle", which contradicts a screenshot showing both toggles.
- **Base is `feat/gcp-integration-docs`, not `main`.** Reviewed as a stacked diff; none of these pages are in production yet, which is why the redirect is belt-and-braces rather than required.
- Assumed the three retained "check the Cloud Run logs" mentions are intentional, per the PR description. Verified they refer to collector stdout, not GCP log ingestion.
- Did not run `yarn build`; relied on the PR's report that it passes.

### Onboarding label: skipped

No file has `status == "added"` — the only added paths are two `next.config.js` redirect entries. The one added-then-removed docs file (`collect-logs.mdx`) is a deletion. Per the policy, edits to existing docs do not qualify.

revmag (approved):

crazyuploader (approved):

revmag (approved):

PR #3924, opened 2026-08-09
docs: update invite team members documentation for multi-role support

revmag:
## Description

Updates the Invite Team Member page for the unified invite flow and multi-role support.

- Documents assigning **one or more roles per invitee** (the picker is now multi-select; roles appear as tags and can be deselected individually).
- Renames the column reference to **Roles** and notes the dropdown lists all built-in and custom roles.
- Adds per-row inline email validation (`Invalid email address`) and submission-level callout behavior.
- Documents partial-success handling: the form stays open, failed rows show inline errors, and a `Some invites failed` toast (Members) / `Some invites failed. Check the errors above.` notification (onboarding) appears.
- Updated the `date` frontmatter to today; tightened title and description.

Screenshots are not refreshed: the demo currently lags #12476 and still shows the single-select role UI, so the multi-select modal cannot be captured accurately yet. Prose-only pending demo catch-up.

Source PRs: #11872, #11875, #11873, #12476

Auto-generated by docs-sync pipeline

### DIFF
--- a/data/docs/manage/administrator-guide/iam/invite-team-member.mdx
+++ b/data/docs/manage/administrator-guide/iam/invite-team-member.mdx
@@ -1,7 +1,7 @@
 ---
-date: 2026-05-15
-description: Learn how to invite team members in SigNoz, assign roles, send invites, and manage pending invitations.
-title: Invite Team Member
+date: 2026-08-09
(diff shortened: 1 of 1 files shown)


vikrantgupta25:
This is a very old doc. Will remove this here: https://github.com/SigNoz/platform-pod/issues/2926 and add this change to the new doc.

PR #4002, opened 2026-08-17
docs: document search() in the functions reference

tushar-signoz:
## Pull Request

---

### 📄 Summary

`search()` is a query function users can type in the Logs Explorer filter bar, but it is not documented anywhere. This PR adds it to the **Functions Reference**, next to `hasToken()` and `has()`, which is where a reader goes to look up query-function syntax.

The entry is intentionally self-contained, so it stands alone and does not depend on any other page:

- **Syntax** — `search(term)` and `search(term, scope, ...)`, with examples of the keyless form, a scoped form, multiple scopes, and combining with a field filter.
- **Scopes** — table of the four valid field contexts (`body`, `attribute`, `resource`, `log`) and what each covers, plus the note that attribute/resource **keys** are searched as well as values, and that an invalid scope or a qualified field like `body.message` returns HTTP 400.
- **Matching rules** — case-insensitive, always literal (never regex), substring rather than whole word, and when to quote the term.
- **Query cost** — the advisory warning returned on every call, the per-shard scan-budget rejection, the 5s plan timeout, the three ways to get back under budget, and the lower budget on native-JSON-body logs with the self-hosted `querier.search_max_scan_rows*` options.

Also updates *Important Notes* (`has()` is the body-JSON-only one; `search()` is logs-only), adds a *Common Use Cases* example for "the value could be anywhere", adds the `invalid search scope` error, and adds a *Related* section.

Scope is limited to `data/docs/userguide/functions-reference.mdx` — no nav, redirect, or new-page changes.

#### Drive-by fix

The *Filtering by Tags* code block on `main` was never closed, so the `## Error Messages` heading and the error list below it rendered inside a code block. The closing fence is added.

#### Issues closed by this PR

N/A

---

### ✅ Change Type
_Select all that apply_

- [x] ✨ Feature (new documentation)
- [x] 🐛 Bug fix (unterminated code fence)
- [ ] ♻️ Refactor
- [ ] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

### 🧪 Testing Strategy

- Tests added/updated: not applicable — content-only change
- Manual verification:
  - `yarn check:docs-metadata` + `yarn test:docs-metadata` — pass (30/30)
  - `yarn check:doc-redirects` + `yarn test:doc-redirects` — pass (13/13)
  - All linked routes verified to exist on `main`: `json-logs`, `operate/configuration`, `search-syntax`, `operators-reference`, `search-troubleshooting`, `querying/query-performance`
- Edge cases covered: internal links use absolute `https://signoz.io/...` URLs per the docs link convention; no external links added

---

### ⚠️ Risk & Impact Assessment

- Blast radius: one docs page
- Potential regressions: none — no nav, route, or redirect changes, so no inbound links move
- Rollback plan: revert the commit

---

### 📝 Changelog

| Field | Value |
|------|-------|
| Deployment Type | Cloud / OSS / Enterprise |
| Change Type | Feature |
| Description | Documents the `search()` query function — scopes, matching rules, and scan-budget limits — in the Functions Reference guide. |

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented (none)

---

## 👀 Notes for Reviewers

Split out of #3958. That PR carries the full → free rename plus a standalone `search()` guide page; this one adds `search()` only to the functions page. This branch is based on `main`, but both PRs edit `data/docs/userguide/functions-reference.mdx`, so whichever lands second needs a rebase.

Every string and behavior quoted from the backend was verified against the querier source, with references in the comment below.

Worth a second look on the exact strings quoted from the backend: the advisory warning text, the scan-budget rejection message, the `This query is too broad to plan within 5s` message, and the two `querier.search_max_scan_rows*` defaults.

---

### DIFF
--- a/data/docs/userguide/functions-reference.mdx
+++ b/data/docs/userguide/functions-reference.mdx
@@ -1,18 +1,111 @@
 ---
-date: 2026-07-27
+date: 2026-08-17
 
 title: Functions Reference Guide
-description: Reference guide for SigNoz query functions for arrays and token matching. Learn syntax and usage examples.
+description: Reference guide for SigNoz query functions. Use search() for text across every log field, hasToken() for token matching, and has() for arrays.
 doc_type: reference
 ---
 
-This guide covers functions available for querying array fields in SigNoz.
+This guide covers functions available in SigNoz query expressions.
 
 <Admonition type="info">
-Functions are currently supported only for JSON body searches.
+Except for `search()` and `hasToken()`, these functions are supported only for JSON body searches.
 </Admonition>
 
 ## Available Functions
+
+### search() Function
+
+Runs a case-insensitive text search for a term across every field of a log record: the body, attribute keys and values, resource keys and values, and log fields such as severity. Use it when you know what you are looking for but not which field holds it.
+
+**Syntax:**
+```
+search(term)
+search(term, scope, ...)
+```
+
+**Examples:**
+```
+# Every field
+search('timeout')
+
+# Only the log body
+search('timeout', body)
+
+# Body and resource attributes
+search('checkout', body, resource)
+
+# Combined with a field filter
+search('timeout') AND service.name = 'checkout'
+```
+
+#### Scopes
+
+Each optional scope narrows the search to one field context, and several scopes search the union of them. Scopes can be written bare or quoted, so `search('checkout', body)` and `search('checkout', 'body')` are equivalent.
+
+| Scope | Fields covered |
+| --- | --- |
+| `body` | The log body |
+| `attribute` | Keys and values of string, number, and boolean attributes |
+| `resource` | Keys and values of resource attributes |
+| `log` | `severity_text`, `trace_id`, `span_id` |
+
+Keys are searched as well as values, so `search('tenant')` also matches a log that carries a `tenant` attribute whatever its value.
+
+<Admonition type="warning">
+A scope must be one of the four field contexts above. An unknown word or a qualified field such as `body.message` is rejected with an HTTP 400 error. To match inside a specific body key, use a field filter instead.
+
+```
+# Invalid: not a field context
+search('checkout', bogus)
+
+# Invalid: a qualified field is not a scope
+search('checkout', body.message)
+
+# Correct: filter the field directly
+body.message CONTAINS 'checkout'
+```
+</Admonition>
+
+#### Matching Rules
+
+- **Case-insensitive** - `search('CHECKOUT')` and `search('checkout')` return the same logs.
+- **Literal, never regex** - regex metacharacters match as written, so `search('price: $99.99')` looks for that exact text rather than treating `$` as an end anchor.
+- **Substring, not whole word** - `search('uuid')` matches `abcuuid123xyz`. For whole-token matching, use [hasToken()](https://signoz.io/docs/userguide/functions-reference/#hastoken-function).
+
+Quote the term. An unquoted term is accepted only when it is a single bare word, number, or boolean, so anything containing a space or punctuation must be quoted:
+
+```
+# Correct
+search('connection refused')
+
+# Wrong: unquoted multi-word text is a syntax error
+search(connection refused)
+```
+
+#### Query Cost
+
+`search()` reads every searchable column instead of a single indexed one, so it is far more expensive than a field filter. SigNoz applies two guardrails.
+
+Every call returns an advisory warning alongside its results. The results are still correct; the warning is a nudge to switch to a field filter once you know where the term lives:
+
+```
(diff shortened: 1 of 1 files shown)


data/docs/userguide/functions-reference.mdx
```diff
@@ -1,19 +1,112 @@
...
+# Body and resource attributes
+search('checkout', body, resource)
+
+# Combined with a field filter
+search('timeout') AND service.name = 'checkout'
+```
+
+#### Scopes
+
+Each optional scope narrows the search to one field context, and several scopes search the union of them. Scopes can be written bare or quoted, so `search('checkout', body)` and `search('checkout', 'body')` are equivalent.
```
krsoninikhil: Let's use the context instead of scope since that's what we use in the implementation as well and scope can be confused with scope column if we want to add the support for that here.

data/docs/userguide/functions-reference.mdx
```diff
@@ -1,19 +1,112 @@
 ---
-date: 2026-07-27
+date: 2026-08-17
 
 title: Functions Reference Guide
-description: Reference guide for SigNoz query functions for arrays and token matching. Learn syntax and usage examples.
+description: Reference guide for SigNoz query functions. Use search() for text across every log field, hasToken() for token matching, and has() for arrays.
```
krsoninikhil: We don't support `hasAny` / `hasAll`?

data/docs/userguide/functions-reference.mdx
```diff
@@ -1,19 +1,112 @@
...
+
+# Wrong: unquoted multi-word text is a syntax error
+search(connection refused)
+```
+
+#### Query Cost
+
+`search()` reads every searchable column instead of a single indexed one, so it is far more expensive than a field filter. SigNoz applies two guardrails.
+
+Every call returns an advisory warning alongside its results. The results are still correct; the warning is a nudge to switch to a field filter once you know where the term lives:
```
krsoninikhil, line 90: Let's also add a warning about not using this is dashboard.

krsoninikhil (changes requested):

data/docs/userguide/functions-reference.mdx
tushar-signoz: Yes good catch, dashboards and alerts

data/docs/userguide/functions-reference.mdx
tushar-signoz: Done in e6f1165 — headings, prose, the syntax block and the table now say "field context". The one remaining "scope" is the quoted `invalid search scope` error, kept verbatim because that is the string the querier emits.

Worth flagging that the clash you're guarding against is already live, not just hypothetical:

- `FieldContextFromText` resolves against the shared `fieldContexts` map (`pkg/types/telemetrytypes/field_context.go:67`), which includes `"scope"`, plus `span`, `trace`, `metric`, `event` and the `tag`/`point`/`logfield` aliases. So the visitor accepts all of them.
- `searchColumns()` (`pkg/telemetryschema/logstelemetryschema/field_mapper.go:641`) only has cases for `log`, `body`, `attribute`, `resource`. Anything else hits `default`, which returns the union of every column.

Net effect today: `search('x', scope)` neither narrows nor errors, it silently searches every field. Should the querier reject a context it can't narrow to? Happy to document whichever way you want it to behave.

data/docs/userguide/functions-reference.mdx
tushar-signoz: We do support them — `HASANY`/`HASALL` are grammar tokens and are named in `where_clause_visitor.go:700-705`; they're body-JSON-only, same as `has()`. They had just never been documented on this page.

Added in e6f1165: `hasAny()` matches when at least one value is present, `hasAll()` only when every one is, with both argument forms (`hasAny(k, ['a','b'])` and `hasAny(k, 'a', 'b')`) and the HTTP 400 you get for mixing them. The frontmatter description and the Important Notes now list all three array functions.

data/docs/userguide/functions-reference.mdx
tushar-signoz: Added in e6f1165, covering alerts as well per Tushar.

One thing that shaped the wording: nothing in the backend actually blocks this. The cost guard is attached for every request type (`statement_builder.go:414/581/709`), so a saved panel or rule isn't rejected up front — it repeats the full scan on every refresh or evaluation under the same per-shard budget, and starts failing outright once a wider range pushes the estimate over. So the admonition frames it as "don't save it, use a field filter once you know the field" rather than "unsupported". Let me know if you'd rather it read as a hard restriction.

krsoninikhil (approved):

revmag (approved):

PR #4022, opened 2026-08-19
fix(revalidate): on demand revalidation

manika-signoz:
## Pull Request

---

Sync-triggered cache invalidations caused parallel fetches, which sent multiple requests to CMS.

1. cms fetch wrapper adds a 15s timeout and logging to call sites.
2. serve stale content instead of 504 - changed revalidateTag(tag, 'default') to revalidateTag(tag, 'max')
3. limit list-tag invalida to add/delete only instead of all changes
4. add caching for release changelog with 5 min ttl

#### Screenshots / Screen Recordings (if applicable)

NA

N/A 

---

### ✅ Change Type
_Select all that apply_

- [ ] ✨ Feature
- [x] 🐛 Bug fix
- [ ] ♻️ Refactor
- [x] 🛠️ Infra / Tooling
- [ ] 🧪 Test-only

---

#### Root Cause
- Content-sync revalidation invalidated `*-list` tags on every sync (including edit-only), invalidating the all data cache. 
- Changelog fetches were `no-store`, hitting the CMS on every request.

#### Fix Strategy
Reduce list tag invalidation frequency, reduce long awaited calls.
---

- Tests added/updated:
  - `tests/sync/pr-file-classifier.test.js` - added-file classification while PR is open
  - `tests/sync/revalidate-args.test.js` - new CLI args; list-tag invalidation scoping (edit-only changes skip list invalidation)
- Manual verification:
  - Preview deploy: manual `POST /api/revalidate` (paths + tags) → `/rss/` shows `x-vercel-cache: REVALIDATED`; `[cms-fetch]` logs confirm timeout wrapper and staging-CMS wiring
  - Local production builds against staging CMS pass; route table and prerender manifest inspected for sitemap/RSS routes
- Edge cases covered: legacy sync callers without `--added-files` (fallback to invalidate on any change), non-OK page responses still throw (partial response is never cached)

---

- Blast radius: all CMS-backed rendering (docs, blog, comparisons, guides, faqs, case studies, changelog), `/api/revalidate` consumers, content-sync workflow.
- Potential regressions:
  - 15s timeout could abort slow CMS responses.
  - Edit-only syncs no longer refresh list ordering metadata that depends on edited fields (e.g. updated dates in listing pages) until the next membership change or 24h TTL - accepted trade-off.
  - Changelog content stays stale for 5 min.
- Rollback plan: revert the PR.

---

### 📋 Checklist
- [x] Manually tested
- [x] Breaking changes documented

---


utils/cmsFetch.ts
```diff
@@ -0,0 +1,31 @@
...
+export async function cmsFetch(url: string, init: RequestInit = {}): Promise<Response> {
+  const method = init.method ?? 'GET'
+  const startedAt = Date.now()
+  console.log(`[cms-fetch] → ${method} ${url}`)
+
+  try {
+    const response = await fetch(url, {
+      ...init,
+      signal: AbortSignal.timeout(CMS_FETCH_TIMEOUT_MS),
+    })
```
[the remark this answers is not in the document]
manika-signoz: will remove later if needed

app/(site)/docs/sitemap.ts
```diff
@@ -4,6 +4,8 @@ import { toSitemapDateOnly } from 'utils/sitemapXml'
 import { resolveLatestDate } from '@/utils/dateUtils'
 import { fetchAllDocsForPage } from '@/utils/cachedData'
```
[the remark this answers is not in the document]
manika-signoz: fine

utils/strapi.ts
```diff
@@ -425,38 +453,54 @@ export const fetchMDXContentByPath = async (
...
+          }
+
+          pageFetch = runWithConcurrency(pageTasks, CMS_PAGE_CONCURRENCY).finally(() => {
+            pageFetches.delete(pageFetchKey)
+          })
+          pageFetches.set(pageFetchKey, pageFetch)
         }
 
-        const pagesResults = await Promise.allSettled(pagePromises)
+        const pagesResults = await pageFetch
```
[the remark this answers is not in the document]
manika-signoz: fetchAll should throw if any page fetch rejects so partial results are not cached, fixed

utils/strapi.ts
```diff
@@ -327,6 +330,32 @@ const singleContentPopulateByCollection: Record<string, Record<string, unknown>>
...
+        results[index] = { status: 'rejected', reason }
+      }
+    }
+  }
+
+  await Promise.all(Array.from({ length: Math.min(limit, tasks.length) }, worker))
+  return results
+}
+
+const pageFetches = new Map<string, Promise<SettledResult<MDXContentApiResponse>[]>>()
```
H4ad, line 352: Would be better to use a lib to do this, such as bluebird.map

Because you don't clear this, it will likely cause a leak

utils/cmsFetch.ts
```diff
@@ -0,0 +1,31 @@
+import { CMS_FETCH_TIMEOUT_MS } from '@/constants/cache'
+
+const errorName = (error: unknown): string => (error instanceof Error ? error.name : '')
+
+export async function cmsFetch(url: string, init: RequestInit = {}): Promise<Response> {
```
H4ad: You could clean/remove all API_URL concats and only leave the path since you know this will always be CMS URL for this function

utils/strapi.ts
```diff
@@ -425,45 +453,66 @@ export const fetchMDXContentByPath = async (
...
-            arrayFormat: 'repeat',
+        const pageFetchKey = [
+          collectionName,
+          deployment_status ?? '',
+          fields?.join(',') ?? '',
+          totalPages,
+        ].join('|')
+
+        let pageFetch = pageFetches.get(pageFetchKey)
+        if (!pageFetch) {
```
H4ad, line 461: This adds two levels of cache, why?

Also, how do you invalidate this first cache when new item in the page since this is in-memory?

manika-signoz: this is not a cache layer, caching and invalidate is only for data cache - the tagged request calls, this is just to avoid hitting same request from 2 entry points in the same instance so the cms does not get multiple requests at once in one instance, and we clear this in finally, the data cache invalidates via revalidateTag

utils/strapi.ts
manika-signoz: i cleared the entries in finally

utils/cmsFetch.ts
manika-signoz: done

app/(site)/api/revalidate/route.ts
```diff
@@ -14,6 +14,36 @@ interface RevalidationResult {
...
 }
 
+const LIST_TAG_DEPENDENT_PATHS: Record<string, string[]> = {
+  'docs-list': ['/docs/sitemap.xml'],
+  'blogs-list': ['/blogs/sitemap.xml', '/rss', '/rss/[...segments]'],
+  'comparisons-list': ['/blogs/sitemap.xml'],
+  'guides-list': ['/blogs/sitemap.xml'],
+  'faqs-list': ['/blogs/sitemap.xml'],
+  'case-studies-list': ['/blogs/sitemap.xml'],
+  'opentelemetries-list': ['/blogs/sitemap.xml'],
```
H4ad, line 24: All of them should invalidate /rss since they are being used by /rss as well, right?

app/(site)/api/revalidate/route.ts
```diff
@@ -14,6 +14,36 @@ interface RevalidationResult {
   timestamp: string
 }
 
+const LIST_TAG_DEPENDENT_PATHS: Record<string, string[]> = {
+  'docs-list': ['/docs/sitemap.xml'],
+  'blogs-list': ['/blogs/sitemap.xml', '/rss', '/rss/[...segments]'],
```
H4ad: `'/rss/[...segments]'`

This seems to be no-op by claude, 

```
Traced end to end in next@16.2.11:

1. Purge side — revalidatePath(p, 'page') builds tag _N_T_ + path + /page:
node_modules/next/dist/server/web/spec-extension/revalidate.js:83-91 → emits _N_T_/rss/[...segments]/page.
2. Write side — route handler tags come from getImplicitTags(this.definition.page, req.nextUrl.pathname, null):
node_modules/next/dist/server/route-modules/app-route/module.js:534.
3. definition.page always ends in /route for app routes — matcher keys are filtered by isAppRouteRoute:
route-matcher-providers/app-route-route-matcher-provider.js:24 + lib/is-app-route-route.js:11 (route.endsWith('/route')).
So page = /rss/[...segments]/route.
4. Derived tags (lib/implicit-tags.js:15-34) for that page:
_N_T_/layout
_N_T_/rss/layout
_N_T_/rss/[...segments]/layout
_N_T_/rss/[...segments]/route
plus pathname tag _N_T_/rss/tags/<slug>.

_N_T_/rss/[...segments]/page is not in that set. Zero match → tag feeds never purge. Suffix page only exists for App Pages, never Route Handlers.

Also confirms the plain /rss entry does work — via the pathname tag _N_T_/rss, not a derived tag.

Fix stands: revalidatePath('/rss', 'layout') → _N_T_/rss/layout, which is in the derived set of both /rss/route and /rss/[...segments]/route. One call, covers everything under /rss. Then drop both /rss entries and the p.includes('[') heuristic.
```

app/(site)/api/revalidate/route.ts
```diff
@@ -76,13 +106,15 @@ export async function POST(request: NextRequest) {
...
     if (revalidateAll) {
       revalidatePath('/', 'layout')
-      revalidateTag('mdx-content-list', 'default')
-      revalidateTag('comparisons-list', 'default')
-      revalidateTag('guides-list', 'default')
-      revalidateTag('blogs-list', 'default')
-      revalidateTag('docs-list', 'default')
-      revalidateTag('docs-side-nav', 'default')
-      revalidateTag('listicles', 'default')
+      revalidateTag('mdx-content-list', 'max')
```
H4ad, line 109: nit: for future PR, move these tags to a enum/file, to centralize these values to avoid typo mistakes

app/(site)/api/revalidate/route.ts
manika-signoz: yes, list tag invalidation accomplishes the goal, tested this manually on staging

app/(site)/api/revalidate/route.ts
manika-signoz: fixed

H4ad (approved):

scripts/cms-sync/pr-file-classifier.js
manika-signoz, line 120: .

makeavish (approved):

PR #3984, opened 2026-08-13
feat(docs): Add observability and OpenTelemetry glossary

revmag:
## What

Adds **`/docs/glossary`** — 73 alphabetical terms with an A–Z letter index, covering OpenTelemetry, observability practice, and SigNoz specifics. Every term is its own `###` heading, so each has a stable anchor that can be linked directly (e.g. `.../docs/glossary/#identifying-attribute`).

Also corrects the opening of **`metrics-management/reducing-costs`**, which stated the billing model incorrectly.

## Preview

- **Glossary** — https://staging.signoz.io/docs/glossary/
- **Metrics billing (rewritten)** — https://staging.signoz.io/docs/metrics-management/reducing-costs/
- Anchor example for support replies — https://staging.signoz.io/docs/glossary/#identifying-attribute

## Why

Two gaps that keep coming up when someone tries to model and attribute metrics cost:

1. **No canonical vocabulary.** "Identifying attribute" is used in our billing doc but defined nowhere, and there's no documented set. Without a canonical term, people invent their own — "unique label set" being the common one — and then can't map it back to our docs.
2. **The billing doc stated the model wrong.** It opened with "Metrics billing is based on **cardinality** - the number of unique time series created by your metrics." Metrics are billed per sample ingested. Cardinality is the primary cost driver by arithmetic, not by pricing unit — and stating it as the unit reads as contradicting the fact that we don't bill per time series, which is a deliberate differentiator.

## Glossary contents

**Metrics / cost cluster** — cross-linked so a reader landing on one can't miss the others:

| Term | Substance |
|---|---|
| Sample | The billing unit, $0.10/M. Gauge/sum = 1; histogram = (buckets × 1) + 4 for `.sum`/`.count`/`.min`/`.max` |
| Time series | Real for storage and querying (fingerprint, label set stored once, Metrics Explorer counts) — explicitly **not** a billing unit |
| Identifying attribute | With the worked example: one label set + `http.route` at 10 distinct values → 10 series, 10 datapoints per interval |
| Non-identifying attribute | `host.type` given `host.id`; EC2 tags from the AWS resource detector add series identity but no series and no cost |
| Unique label set | Included as its own entry, pointing at the canonical terms — it's the phrase people reach for when none is documented |
| Fingerprint | Hash over metric name, temporality, resource attrs, scope attrs, datapoint attrs; a storage/aggregation concept, never billed |
| Churn | Free — a replaced series sends the same number of samples |
[...]

PR #4052, opened 2026-08-25
feat(docs): Wire glossary terms into docs as hover tooltips

revmag:
Stacked on #3984 — based on `docs-observability-glossary`, because the tooltips link to glossary anchors that only exist on that branch. GitHub will retarget this to `main` automatically once #3984 merges.

## What

Brings the blog's `<Tooltip>` treatment to docs. A jargon term keeps its place in the sentence, gains a dashed underline, and on hover shows a one-sentence definition plus a link to its glossary entry.

Five terms, placed by hand, first occurrence per page:

| Page | Term |
|---|---|
| `metrics-management/volume-control` | cardinality |
| `metrics-management/dropping-metric-labels` | cardinality |
| `metrics-management/aggregated-metric-data` | cardinality |
| `metrics-management/metrics-explorer` | cardinality, temporality |

Rendered output, with no `target="_blank"` and no stray space:

```html
<span class="relative inline-block">
  <a href="https://signoz.io/docs/glossary/#cardinality"
     class="cursor-pointer no-underline border-b border-dashed …">cardinality</a>
</span>
```

## Where terms were deliberately skipped

Curation matters more than coverage here. `service` appears 4,859 times in `data/docs` and `trace` 2,253 — tagging everything would put dashed underlines on several words per paragraph. Beyond that:

- **The page already defines the term itself.** `dropping-metric-labels` has `- **Time Series**: A unique combination of metric name + label set`, so a competing glossary definition on the same page would be worse than none.
- **The only occurrences aren't prose** — headings, table cells, JSX attributes such as `Figure` `alt=`/`caption=`, or UI element names ("the **Time Series** view"). `metrics-explorer` mentions "time series" 19 times and not one is a safe target.
- **Frontmatter.** The first match on several pages is in `title`/`description`.

## Two prerequisites, not polish

**1. `Tooltip` had no agent-markdown stub.** Docs render twice: once as the page, once as plain text at `/api/docs-markdown/…` for AI tools. Without a stub, `tests/agent-markdown-component-coverage.test.js` fails for any docs page using the component, and the generic fallback emitted:

```
A metric with high

Component: Tooltip

costs more.
```

The term deleted, one sentence split into three fragments, for every AI reader. Now:

```
Volume control lets you reduce the stored volume and [cardinality](https://signoz.io/docs/glossary/#cardinality) of a metric by …
```

**2. `components/ui/Tooltip.tsx` forced `target="_blank"` on every link** — wrong for an internal glossary anchor. Fixed, plus four other defects found while testing:

- External links keep `target`/`rel` (now `noopener noreferrer nofollow`, matching the docs link convention); `signoz.io`, root-relative and `#` links stay in the same tab.
- Removed the stray `&nbsp;` after the trigger. It rendered a double space in prose and in Copy Markdown — visible today on the blog as "metrics  work in time-series".
- The no-link trigger was a plain `<span>` with no `tabIndex`, so keyboard users could never reach the definition. Now focusable, with `aria-describedby` and Escape-to-dismiss.
- The hover card was a `<div>` inside a `<span>` inside a `<p>`. Invalid nesting, and a hydration-mismatch risk now that these sit in docs prose. Now a `<span class="block">`.
- The pending hide timeout wasn't cleared on unmount.

These are shared with the blog, so its 15 existing tooltips get the same fixes.

## Verification

- `agent-markdown-component-coverage` and `agent-markdown-stubs` tests pass
- All four pages render 200 locally with the expected 5 triggers
- Agent markdown checked on all four: **zero** `Component: Tooltip`, terms present as links
- Rendered anchor confirmed free of `target="_blank"` and `&nbsp;`
- `check:docs-metadata`, `check:doc-redirects`, `check:stale-urls`, `test:stale-urls` (19 pass)
- `yarn lint` 0 errors, no warnings in either changed file; `yarn build` exits 0

## Notes for review

- Only `cardinality` and `temporality` are used so far. That's deliberate for a first pass — worth agreeing the pattern reads well before spreading it.
- This is hand-placed. Auto-injecting from `glossary.mdx` via a remark plugin is the obvious follow-up, but it needs three decisions first: scoping the plugin to docs so it doesn't fight the blog's hand-curated tooltips, fetching the glossary through the CMS path so definitions can't diverge from the glossary page, and revalidating consumer pages on a glossary edit the way `findDocsPathsForListicle` already does for Listicles. Happy to open that separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/components/ui/Tooltip.tsx
+++ b/components/ui/Tooltip.tsx
@@ -29,6 +29,22 @@ const GAP = 8
 // Keep the tooltip below the sticky top navigation, which is about 64px tall.
 const NAV_SAFE_TOP = 72
 
+// Find the nearest ancestor that clips its content (the docs content column uses
+// `overflow-clip`). The tooltip itself is portalled out, so this is used only to
+// keep the card horizontally inside the reading column instead of sprawling over
+// the sidebar or table of contents.
+function getClipAncestor(element: HTMLElement | null): HTMLElement | null {
+  let node = element?.parentElement ?? null
+  while (node && node !== document.body) {
+    const style = getComputedStyle(node)
+    if ([style.overflow, style.overflowX, style.overflowY].some((value) => value !== 'visible')) {
+      return node
+    }
+    node = node.parentElement
+  }
+  return null
+}
+
 interface Position {
   top: number
   left: number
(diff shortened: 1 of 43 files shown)


components/ui/Tooltip.tsx
```diff
@@ -13,14 +13,23 @@ interface TooltipProps {
   linkText?: string
 }
 
+/** signoz.io links stay in the same tab; only third-party links open a new one. */
+function isInternalHref(href: string) {
```
manika-signoz: .

manika-signoz (approved):

PR #4201, opened 2026-09-18
docs: document Slack attachment options and webhook basic auth

revmag:
## Description

- Document the new Slack notification-channel attachment fields (`color`, `title_link`, `pretext`, `fallback`, `footer`, `fields[]`, `actions[]`) with a reference table and an annotated example payload on the Slack channel page. These are API-only; the UI form does not expose them yet (backend types + generated schema only).
- Clarify that webhook basic auth accepts a username or password independently (neither requires the other), and that a bearer token cannot be combined with basic auth.
- Updated frontmatter `date` on both edited files.

Source PR: SigNoz/signoz#12907


namanv3:
doc not needed. this is for v2 api that isn't released yet as such

